PR #80, opened 2015-01-08
Build: remove dates from copyright notice

arthurvr:
This should have been done in https://github.com/jquery/jquery-migrate/commit/153233281c3ed4256413208e04c091aa1ab36313, but seems like it isn't.


dmethvin:
Not sure how i missed that one, thanks!

PR #81, opened 2015-01-09
Core: Fill in context and selector properties for .live() support

### LINKED ISSUE #79: Move selector and context properties to Migrate for `.live()` support
See https://github.com/jquery/jquery/issues/1908

dmethvin:
Fixes #79

Ref https://github.com/jquery/jquery/issues/1908

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -4,3 +4,11 @@ jQuery.fn.andSelf = function() {
 	migrateWarn("jQuery.fn.andSelf() replaced by jQuery.fn.addBack()");
 	return oldSelf.apply( this, arguments );
 };
+
+var oldFind = jQuery.fn.find;
+jQuery.fn.find = function( selector ) {
+	var ret = oldFind.apply( this, arguments );
+	ret.context = this.context;
+	ret.selector = this.selector ? this.selector + " " + selector : selector;
+	return ret;
+};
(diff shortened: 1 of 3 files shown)


src/traversing.js
```diff
@@ -4,3 +4,11 @@ jQuery.fn.andSelf = function() {
 	migrateWarn("jQuery.fn.andSelf() replaced by jQuery.fn.addBack()");
 	return oldSelf.apply( this, arguments );
 };
+
+var oldFind = jQuery.fn.find;
```
dmethvin: Needs to be moved to the top of the file.

test/core.js
```diff
@@ -47,6 +47,49 @@ test( "jQuery(html) loose rules", function() {
...
+	expect( 18 );
+
+	var test;
+
+	test = jQuery( undefined );
+	equal( test.selector, "", "Empty jQuery Selector" );
+	equal( test.context, undefined, "Empty jQuery Context" );
+
+	test = jQuery( document );
+	equal( test.selector, "", "Document Selector" );
```
dmethvin: I ran the jquery-migrate unit tests against the core PR built locally, by using the "dev" option in the jquery-migrate unit test dropdown. It seemed to pass so we should be good. We should actually land this PR first I'd say, to be sure that it works correctly with older core versions on all supported browsers.

mzgol: OK, I just wanted to get it tested properly. Since you checked it on your own, I'm good. :)

dmethvin:
Landed at 683b9b3ae40e3fe9e83f3dd39d8e5e958975f3fa

Issue #89, opened 2015-02-27
Complete warnings/shims for jQuery 1.8 deprecations


dmethvin:
http://api.jquery.com/category/deprecated/deprecated-1.8/

- [x] Docs in warnings.md for `.andSelf()` and `.size()` (Ref #62)
- [x] `.load()`
- [x] `.unload()`
- [x] `deferred.pipe()`

N.B. Since the `.load()` and `.unload()` shims are just like [`.error()`](https://github.com/jquery/jquery-migrate/blob/683b9b3ae40e3fe9e83f3dd39d8e5e958975f3fa/src/event.js#L41-51) the existing code can just be generalized. None of those need to bubble so `.triggerHandler()` will do the trick.

~~@gibson042 Now that we're going all Promise/A+ on Deferred should we un-deprecate `deferred.pipe` to preserve a way to get the old behavior? If so I can open a docs ticket for that tagged for 3.0.~~

dmethvin:
As discussed in the meeting today, we're leaving `deferred.pipe()` deprecated and should warn for it.

dmethvin:
I'll close once the 3 PRs land

dmethvin:
All related PRs have landed. :fireworks:

PR #93, opened 2015-03-12
Deferred: Deprecate deferred.pipe() method

dmethvin:
Ref #89

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -4,10 +4,18 @@ var oldDeferred = jQuery.Deferred;
 jQuery.Deferred = function() {
 	var deferred = oldDeferred.apply( this, arguments );
 
-	deferred.pipe = function() {
-		migrateWarn( "deferred.pipe() is deprecated" );
-		return deferred.then.apply( this, arguments );
-	};
-	
+	// Don't add this method if the current jQuery doesn't provide it
+	if ( deferred.pipe ) {
+
+		deferred.pipe = function() {
+			migrateWarn( "deferred.pipe() is deprecated" );
+			return deferred.then.apply( this, arguments );
+		};
+
+		deferred.promise().pipe = function() {
+			return deferred.pipe.apply( this, arguments );
+		};
+	}
+
 	return deferred;
 };
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -1,0 +1,13 @@
+
+var oldDeferred = jQuery.Deferred;
+
+jQuery.Deferred = function() {
+	var deferred = oldDeferred.apply( this, arguments );
+
+	deferred.pipe = function() {
+		migrateWarn( "deferred.pipe() is deprecated" );
+		return deferred.then.apply( this, arguments );
+	};
+	
+	return deferred;
+};
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -27,6 +27,7 @@ module.exports = function(grunt) {
 			"src/manipulation.js",
 			"src/event.js",
 			"src/traversing.js",
+			"src/deferred.js",
 			"src/outro.js"
 		],
 		tests: {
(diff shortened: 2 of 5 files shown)


src/deferred.js
```diff
@@ -0,0 +1,13 @@
+
+var oldDeferred = jQuery.Deferred;
+
+jQuery.Deferred = function() {
+	var deferred = oldDeferred.apply( this, arguments );
+
+	deferred.pipe = function() {
```
dcherman: So this will add the `pipe` method to the deferred, but not to the promise - using `.then` or `.promise` would yield a promise that does not have a `pipe` method unless I'm mistaken.

In order to get the method available everywhere it needs to be, this probably needs to look like

```js
function pipe() { ... }

deferred.pipe = deferred.promise().pipe = pipe;
```

dmethvin: Oh yeah, I totally missed that case! I need to go back and double check some of the edge cases since for example a `Deferred` can have a different object applied as its prototypical `promise` object and none of that is being covered here.

dmethvin:
I still feel like I'm missing some cases here, although it doesn't need to be bulletproof. @gibson042 what do you think?

src/deferred.js
```diff
@@ -0,0 +1,21 @@
+
+var oldDeferred = jQuery.Deferred;
+
+jQuery.Deferred = function() {
+	var deferred = oldDeferred.apply( this, arguments );
+
+	// Don't add this method if the current jQuery doesn't provide it
```
dcherman: I think you mean "only add this method if the current jQuery doesn't provide it"

dmethvin: Overriding it for the warning. I didn't want to add `.pipe` if, for example, they were using jQuery 1.5 because `.pipe` wasn't there and maybe they're doing a feature detect or something. I suppose I could always add it but it seemed better to only shim it if the original `Deferred` didn't have it.

But talking to the rubber ducky now I realize that I'm not correctly covering the version 1.6 to 1.8 region where we had `.then` but it didn't return a new promise. In those cases we should use the existing `.pipe` which does return a new promise.

dcherman: I think I was confusing myself;  I was making the false assumption that `.pipe` was being removed in 3.0, but I realize that isn't the case.  What you said about older versions is def true though

jaubourg:
You're probably safer with something like this:

```js
var Deferred = jQuery.Deferred;

jQuery.Deferred = function( func ) {
	var deferred = Deferred();
	var pipe = deferred.pipe;
	if ( pipe ) {
		deferred.pipe = deferred.promise().pipe = function() {
			migrateWarn( "deferred.pipe() is deprecated" );
			return pipe.apply( this, arguments );
		};
	}
	if ( func ) {
		func.call( deferred, deferred );
	}
	return deferred;
};
```

dmethvin:
Thanks @jaubourg! It makes sense to wait until we've shimmed the `.pipe()` to call the function if it was provided, I didn't catch that case.

gibson042:
I like @jaubourg's code and would take it even farther, dropping the `if ( pipe )` in favor of an unconditional definition in accord with the single-minded purpose I've advocated elsewhere ("restore backwards incompatibilities introduced between jQuery 1.11.2 and 3.0.0").

By the way, should I make an issue/PR for that?

dmethvin:
I should be able to get to it in the next day or two. Thoughts on the unit tests?

gibson042:
> Thoughts on the unit tests?

Hmmm, you'll need to be async because of `then` changes, but you'll also want to guarantee that `pipe` is synchronous. Also probably one showing that `dfdRejected.pipe( null, function() { return 1; } )` yields a rejected Deferred (another `then`/`pipe` distinction). Other than that, LGTM.

dmethvin:
Seems like now would be a good time to upgrade QUnit as well then. Let the yak shaving begin!

dmethvin:
Closed in favor of #101

mzgol:
> Closed in favor of #101

It still hasn't been closed. :) I'm closing it then.

PR #94, opened 2015-03-12
Event: .load() and .unload() are deprecated as of 1.8

dmethvin:
Refactored the code to handle .error() since the three have the same form.

Ref #89

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -38,17 +38,25 @@ jQuery.event.remove = function( elem, types, handler, selector, mappedTypes ){
 	eventRemove.call( this, elem, hoverHack( types ) || "", handler, selector, mappedTypes );
 };
 
-jQuery.fn.error = function() {
-	var args = Array.prototype.slice.call( arguments, 0);
-	migrateWarn("jQuery.fn.error() is deprecated");
-	args.splice( 0, 0, "error" );
-	if ( arguments.length ) {
-		return this.bind.apply( this, args );
-	}
-	// error event should not bubble to window, although it does pre-1.7
-	this.triggerHandler.apply( this, args );
-	return this;
-};
+jQuery.each( [ "load", "unload", "error" ], function( _, name ) {
+
+	jQuery.fn[ name ] = function() {
+		var args = Array.prototype.slice.call( arguments, 0 );
+		migrateWarn( "jQuery.fn." + name + "() is deprecated" );
+		args.splice( 0, 0, name );
+		if ( arguments.length ) {
+			return this.bind.apply( this, args );
+		}
+
+		// Use .triggerHandler here because:
+		// - load and unload events don't need to bubble, only applied to window
+		// - error event should not bubble to window, although it does pre-1.7
+		// See http://bugs.jquery.com/ticket/11820
+		this.triggerHandler.apply( this, args );
+		return this;
+	};
+
+});
 
 jQuery.fn.toggle = function( fn, fn2 ) {
 
--- a/src/event.js
+++ b/src/event.js
@@ -4,6 +4,7 @@ var eventAdd = jQuery.event.add,
 	oldToggle = jQuery.fn.toggle,
 	oldLive = jQuery.fn.live,
 	oldDie = jQuery.fn.die,
+	oldLoad = jQuery.fn.load,
 	ajaxEvents = "ajaxStart|ajaxStop|ajaxSend|ajaxComplete|ajaxError|ajaxSuccess",
 	rajaxEvent = new RegExp( "\\b(?:" + ajaxEvents + ")\\b" ),
 	rhoverHack = /(?:^|\s)hover(\.\S+|)\b/,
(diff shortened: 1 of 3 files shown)


warnings.md
```diff
@@ -93,6 +93,13 @@ $.ajax({
 
 **Solution:** Change any use of `$().error(fn)` to `$().on("error", fn)`.
 
+### JQMIGRATE: jQuery.fn.load() is deprecated
+### JQMIGRATE: jQuery.fn.unload() is deprecated
+
+**Cause:** The `$().load()` and `$().unload()` methods attach an "load" and "unload" event, respectively, to an element. They were deprecated in 1.9 to reduce confusion with the AJAX-related `$.load()` method that loads HTML fragments and which has not been deprecated. Note that these two methods are used almost exclusively with a jQuery collection consisting of only the `window` element. Also note that attaching an "unload" or "beforeunload" event on a window via any means can impact performance on some browsers because it disables the document cache (bfcache). For that reason we strongly advise against it.
+
```
timmywil: The ajax-related method is a prototype method: `$().load()`.

src/event.js
```diff
@@ -38,17 +38,25 @@ jQuery.event.remove = function( elem, types, handler, selector, mappedTypes ){
...
-};
+jQuery.each( [ "load", "unload", "error" ], function( _, name ) {
+
+	jQuery.fn[ name ] = function() {
+		var args = Array.prototype.slice.call( arguments, 0 );
+		migrateWarn( "jQuery.fn." + name + "() is deprecated" );
+		args.splice( 0, 0, name );
+		if ( arguments.length ) {
+			return this.bind.apply( this, args );
+		}
```
timmywil: Need to defer to the existing `.load` – to follow the ajax path – when the first argument is a string.

warnings.md
scottgonzalez: `.load()` not `$().load()`, etc.

dmethvin: Yeah I must have been asleep during this one.

Mottie:
I am testing jQuery 3.0.0-alpha1 and I'm not seeing this JQMIRATE message. Instead, in [this demo](http://jsfiddle.net/Mottie/L4fsj5o3/) I am getting the following error:

    Uncaught TypeError: url.indexOf is not a function  jquery-3.0.0-alpha1.js:9212

dmethvin:
There hasn't been a release with this  yet, the PR is still open.

Mottie:
I was just hoping to expedite the merging of this PR since the [jQuery blog](http://blog.jquery.com/2015/07/13/jquery-3-0-and-jquery-compat-3-0-alpha-versions-released/) is now encouraging us to test out jQuery 3.0.0-alpha1 and use the migrate plugin.

dmethvin:
Just realized I need a test case for the delegating to ajax load in the last commit, so this isn't quite ready to go.

src/event.js
```diff
@@ -38,17 +39,34 @@ jQuery.event.remove = function( elem, types, handler, selector, mappedTypes ){
...
+		// technically this could also be the "Anything" arg of the event .load()
+		// which just goes to show why this dumb signature has been deprecated!
+		// jQuery custom builds that exclude the Ajax module justifiably die here.
+		if ( name === "load" && typeof arguments[ 0 ] === "string" ) {
+			return oldLoad.apply( this, arguments );
+		}
+
+		args.splice( 0, 0, name );
+		if ( arguments.length ) {
+			return ( this.bind || this.on ).apply( this, args );
```
mzgol: Don't we have more uses of `bind` like https://github.com/jquery/jquery-migrate/blob/920afc6637645629e8cc22c9b76cd2fdc59e0d96/src/event.js#L46 that would need similar treatment? ditto for `unbind`.

dmethvin: Good point. It was a spur-of-the-moment thought that if someone used a custom build that excluded `.bind()` it would be nice if this still worked. It's worth thinking about whether we want to support those kind of builds though, because if we did all those cases like the one you pointed out should be changed as well.

mzgol: IMO we don't need to. The only use case I can think of is size but if you care about size you should work on getting rid of warnings and Migrate itself afterwards. ;) And only then it's worth caring about custom builds.

gibson042: Given that the new model for Migrate is backwards only (e.g., jQuery 3.0 + Migrate 3.0 ≈ jQuery [1-2].latest), I see no reason not to use _only_ `.on` and `.off`.

mzgol: Ah, I missed the memo. In that case I agree with @gibson042.

djdeo:
hey, guys, i used to load header and footer like this

```html
  <script src="/js/jquery.min.js"></script>
  <script>
  $(document).ready(function() {
    $("header").load("/header.html");
    $("footer").load("/footer.html");
  });
  </script>
```
Since `.load()`  is deprecated , how can I fix this, the `on()` method won't work either.

```js
$(document).ready(function() {
    $("header").on("load", "/header.html");
    $("footer").on("load", "/footer.html");
  });
```

mgol:
@djdeo You can continue to use `load` for this purpose, it's not deprecated. What was deprecated was the alias for `.no('load', handler)`, precisely because it was confusing that it used the same method name as the `load` you use. Now `load` has only one purpose.

djdeo:
Thanks for your explanation, that's a release, so i can just ignore the
console warning, right😁

2018-01-25 17:37 GMT+08:00 Michał Gołębiowski-Owczarek <
notifications@github.com>:

mgol:
@djdeo AFAIK you shouldn't receive a console warning if you use the non-deprecated signature of `load`. If that's not the case, please open a new issue with a test case that will show the warning, preferably on JSFiddle or JS Bin.

djdeo:
I'm using

jQuery v1.11.0

i'll try to make that JSFiddle ASAP

or you can just check my website https://macrorit.com and the warning
shows in the Console, that might be more accurate


2018-01-29 18:01 GMT+08:00 Michał Gołębiowski-Owczarek <
notifications@github.com>:

mgol:
@djdeo A test case would be better, we can't debug an entire third party website.

djdeo:
OK, will do that later

2018-02-12 1:06 GMT+08:00 Michał Gołębiowski-Owczarek <
notifications@github.com>:

djdeo:
Hi, Michal, I created a demo page today, and tested the .load() event, it
works fine on both v1.11 and the late version, I think there might be sth
wrong with my site, thanks for your support and good luck

2018-02-16 23:00 GMT+08:00 jian deng <deoseoer@gmail.com>:

PR #95, opened 2015-03-12
Core: jQuery("#") is bad syntax

### LINKED ISSUE #85: jQuery(“#”) now throws
As of jQuery 3.0, `jQuery("#")` throws an error since it's an invalid selector, rather than the previous behavior of silently returning an empty set. Need a warning about this and to have it return the aforementioned empty set.

dmethvin:
Closes #85


timmywil:
:+1:

PR #99, opened 2015-03-26
Prepare for npm publish

wbinnssmith:
This:

* Adds a `main` field to the `package.json` pointing to the generated dist copy of jquery-migrate.
* Adds an `npm prepublish` step to run grunt before publishing to npm.
* Moves the jquery dependency to `devDependencies`. jQuery is never directly `require`d by the code, and is implicitly assumed to exist (see [backbone's package.json discussion](https://github.com/jashkenas/backbone/issues/2997#issuecomment-35131774) for more info)

This is deliberately unopinionated about CommonJS. I'd like to solve that issue in a separate pull request, but at least now jquery-migrate can be used with npm and either a script tag or a `jQuery` in the global scope. I'd like to implement something similar to [jQuery's wrapper](https://github.com/jquery/jquery/blob/master/src/intro.js), but we should decide what the API should look like. For example, it could remain the same (ie `require('jquery-migrate')`) and bring jquery in as a formal dependency and patch that, but npm's file tree is not flat for the time being (this is going to change in 3.x), and we might end up patching a different jQuery than the user was expecting.

An alternate API would be to use `require('jquery-migrate')(jQuery)`, passing in a reference to the desired jQuery to patch into a factory function. This would clear up the above issue but makes CommonJS somewhat more of a special case. Then there's the handling of a lack of a `window` object that jQuery handles, and there's a question of whether we should handle that too.

I imagine jQuery plugins face these same issues, so I'd love to hear if you have any experience or thoughts on this.

Thanks!

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,8 @@
+.project
+.settings
+*~
+*.diff
+*.patch
+/*.html
+.DS_Store
+node_modules
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "jquery-migrate",
   "title": "jQuery Migrate",
   "description": "Migrate older jQuery code to jQuery 1.9+",
+  "main": "dist/jquery-migrate.js",
   "version": "1.2.2-pre",
   "homepage": "https://github.com/jquery/jquery-migrate",
   "author": {
@@ -21,10 +22,8 @@
       "url": "https://github.com/jquery/jquery-migrate/blob/master/LICENSE.txt"
     }
   ],
-  "dependencies": {
-    "jquery": "^2.1.3"
-  },
   "devDependencies": {
+    "jquery": "^2.1.3",
     "grunt-git-authors": "3.0.0",
     "grunt-contrib-concat": "0.5.1",
     "grunt-contrib-watch": "0.6.1",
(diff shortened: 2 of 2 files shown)


dmethvin:
Hey @wbinnssmith the PR is [failing CI](http://contribute.jquery.org/CLA/status/?repo=jquery-migrate&sha=fea477917080e6127ba55bc0c9e4b7a0fd1d71bb), it doesn't look like you signed the CLA yet.

wbinnssmith:
I actually did submit the form. Is it handled manually or should I see an
update immediately after signing?

arthurvr:
You signed it, I just confirmed that. Thanks! The bot only runs on a push, it will hopefully be okay next time you push to your branch.

wbinnssmith:
Just pushed again and all looks good :smile:

wbinnssmith:
Let me know if there's anything I can do to help out. Would love to see
this on npm.

wbinnssmith:
Thanks! Would you mind running a quick npm publish to get this onto the
registry?

dmethvin:
I'm not sure it's ready to release, there was a Testswarm failure I need to check into and also I wasn't planning on publishing a new release until jQuery 3.0.0 shipped. So it might be better to cherry-pick this into the last release and publish that.

wbinnssmith:
I just published the latest stable version of jquery-migrate (1.2.1), with this fix backported, [to npm](https://www.npmjs.com/package/jquery-migrate) and added you as an owner.

dmethvin:
Wow, sorry, I just saw your notification! Actually, that is fine with me, thanks!

PR #101, opened 2015-04-06
Deferred: Deprecate deferred.pipe(), isResolved(), isRejected()

### LINKED ISSUE #97: isResovled() function
Hi,

In Jahia 6.6 CMS the jquery version used is v1.6.2. And there is a jquery function isResolved() which is depracated and removed after Jquery 1.8. 

Please see
http://api.jquery.com/deferred.isresolved/

Is this function is added into the migrate library ? If not, it could be added ? 

Good Luck. 
Thanks.

dmethvin:
Ref #89
Fixes #97

This combines all the Deferred shims and uses the actual unit tests from jQuery for .pipe().

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -1,0 +1,63 @@
+
+var oldDeferred = jQuery.Deferred,
+	tuples = [
+		// action, add listener, callbacks, .then handlers, final state
+		[ "resolve", "done", jQuery.Callbacks("once memory"),
+			jQuery.Callbacks("once memory"), "resolved" ],
+		[ "reject", "fail", jQuery.Callbacks("once memory"),
+			jQuery.Callbacks("once memory"), "rejected" ],
+		[ "notify", "progress", jQuery.Callbacks("memory"),
+			jQuery.Callbacks("memory") ]
+	];
+
+jQuery.Deferred = function( func ) {
+    var deferred = oldDeferred(),
+		promise = deferred.promise();
+
+	deferred.pipe = promise.pipe =  function( /* fnDone, fnFail, fnProgress */ ) {
+		var fns = arguments;
+
+		migrateWarn( "deferred.pipe() is deprecated" );
+
+		return jQuery.Deferred(function( newDefer ) {
+			jQuery.each( tuples, function( i, tuple ) {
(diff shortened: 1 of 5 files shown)


dmethvin:
@gibson042 and @jaubourg can you take a look? This PR takes a different approach, always patches `.pipe()` rather than doing it conditionally, and moves over the existing unit tests for `.pipe()`. My goal is to not require any other work here if we remove `.pipe()` in a later jQuery version.

src/deferred.js
```diff
@@ -0,0 +1,63 @@
...
+			jQuery.Callbacks("once memory"), "rejected" ],
+		[ "notify", "progress", jQuery.Callbacks("memory"),
+			jQuery.Callbacks("memory") ]
+	];
+
+jQuery.Deferred = function( func ) {
+    var deferred = oldDeferred(),
+		promise = deferred.promise();
+
+	deferred.pipe = promise.pipe =  function( /* fnDone, fnFail, fnProgress */ ) {
```
gibson042: There's an extra space after the second "=", and the indentation looks wrong from here down.

warnings.md
```diff
@@ -148,6 +148,19 @@ $(document).ajaxStart(function(){ $("#status").text("Ajax started"); });
 
 **Solution**: Boolean properties should generally not be passed to `$().attr` at all; replace with `$().prop` unless you truly intend to update the underlying HTML *attribute*.
 
+### JQMIGRATE: deferred.pipe() is deprecated
+
+**Cause**: The `.pipe()` method on a `jQuery.Deferred` object was deprecated as of jQuery 1.8, when the `.then()` method was changed to perform the same function.
+
+**Solution**: Change all occurrences of `.pipe()` to `.then()`.
```
gibson042: I'd add some text like "and ensure that you aren't relying on context/state propagation or synchronous callback invocation (which were dropped from `.then()` for Promises/A+ interoperability as of jQuery 3.0)".

gibson042:
> This PR takes a different approach, always patches `.pipe()` rather than doing it conditionally, and moves over the existing unit tests for `.pipe()`. My goal is to not require any other work here if we remove `.pipe()` in a later jQuery version.

I wholeheartedly support all of the above. Minor comments provided, but the overall approach is solid.

PR #103, opened 2015-04-24
CSS: Add jQuery.swap and warnings

### LINKED ISSUE #100: Add jQuery.swap
https://github.com/jquery/jquery/pull/2182

dmethvin:
Fixes #100. 

I had to do some unusual things here to monitor the internal calls and not yell about them. Any better ideas on how to do that? AFAICT only these three hooks have been used back to jQuery 1.6.4 which is all Migrate supports.

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -1,0 +1,41 @@
+
+var internalSwapCall = false;
+
+// If this version of jQuery has .swap(), don't false-alarm on internal uses
+if ( jQuery.swap ) {
+	jQuery.each( [ "height", "width", "reliableMarginRight" ], function( _, name ) {
+		var oldHook = jQuery.cssHooks[ name ] && jQuery.cssHooks[ name ].get;
+
+		if ( oldHook ) {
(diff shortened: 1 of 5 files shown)


mzgol:
Clever. :)

src/css.js
```diff
@@ -0,0 +1,41 @@
...
+
+// If this version of jQuery has .swap(), don't false-alarm on internal uses
+if ( jQuery.swap ) {
+	jQuery.each( [ "height", "width", "reliableMarginRight" ], function( _, name ) {
+		var oldHook = jQuery.cssHooks[ name ] && jQuery.cssHooks[ name ].get;
+
+		if ( oldHook ) {
+			jQuery.cssHooks[ name ].get = function() {
+				internalSwapCall = true;
+				oldHook.apply( this, arguments );
```
mzgol: Don't we need to preserve the return value?

dmethvin: Yep, good catch! Unit test should have caught that.

PR #109, opened 2015-05-22
Update attributes.js

mattbrundage:
Diction. "May" connotes permission. "Might" connotes a possibility. (http://english.stackexchange.com/questions/7/may-might-whats-the-right-context)

### DIFF
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -71,7 +71,7 @@ jQuery.attr = function( elem, name, value, pass ) {
 
 		// Warn only for attributes that can remain distinct from their properties post-1.9
 		if ( ruseDefault.test( lowerName ) ) {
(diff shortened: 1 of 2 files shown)


dmethvin:
Thanks, can you update the docs warning message [here](https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnattrselected-may-use-property-instead-of-attribute) as well?

mattbrundage:
Done.

timmywil:
The second answer on that question is more correct. Might and may can be used interchangeable in the present tense, so this PR isn't necessary.

timmywil:
On the other hand, I like might better.

dmethvin:
Thanks, sorry about the delay!

Issue #96, opened 2015-03-16
Detects IE 11 as Mozilla


avodonosov:
In IE 11 `navigator.userAgent === "Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; Media Center PC 6.0; .NET4.0C; .NET4.0E; rv:11.0) like Gecko"`

Therefore `$.browser = {mozilla: true, version: "11.0"}`

dmethvin:
The browser sniffing code in Migrate is the same as the code in jQuery 1.8 from before it was removed. The reason it was removed is exactly this kind of thing, it's not a reliable thing to do and not good practice. So as long as Migrate is reporting the browser incorrectly the same way as jQuery 1.8 is reporting the browser, it is correct.

dmethvin:
See #47 .

avodonosov:
FYI, this "bad practice" dogma doesn't help when there is just no way to detect the needed feature.

dmethvin:
As #47 says, you can easily detect this yourself if you really need it.

PR #119, opened 2015-10-22
Support for IE11

airinterface:
From IE11, user agent doesn't have msie keyword.
This commit is for 'Trident' engine to be identified as IE.


dmethvin:
See #96.

PR #123, opened 2015-11-20
Event: Warn on `ready` event

### LINKED ISSUE #106: Warn on the "ready" event
https://github.com/jquery/jquery/issues/2264

dmethvin:
Fixes #106

We just need to warn here since the event isn't being removed until 3.0.


mzgol:
LGTM

PR #124, opened 2015-11-21
Build: Add jQuery.migrateVersion property

### LINKED ISSUE #90: Version attribute
Add a version attribute as most of jQuery plugins has.
In this case I'd suggest adding something like:
jQuery.fn.migrateVersion : "X.X,X"
Thanks

dmethvin:
Fixes #90 

Needs some review but I think it should be close.


timmywil:
I don't see `src/version.js`. Did you forget a `git add`?

dmethvin:
Whoops! Added.

dmethvin:
Fixed by e638aea0e4a036c4303af3c313389e45235cbc1f

PR #142, opened 2016-02-04
Core: Warn about attribute selectors that have an unquoted hash

### LINKED ISSUE #140: Warn about unqoted attribute selectors containing #
We've removed unofficial "support" for selectors with unquoted attribute values where it has to be quoted, like `a[href=#]` in jQuery 2.2.0/1.12.0. It turns out many people relied on that as we get a lot of bug reports about this change.

Migrate should warn about it.

dmethvin:
Fixes #140

Although technically a selector can be used in a lot of places in the API, all our reports in #140 are coming from a simple pattern that this does warn about. Since we don't actually fix the problem it doesn't seem that critical to be thorough with detecting it either.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -3,6 +3,7 @@ var matched, browser,
 	oldInit = jQuery.fn.init,
 	oldParseJSON = jQuery.parseJSON,
 	rspaceAngle = /^\s*</,
+	rattrHash = /\[\s*\w+\s*[~|^$*]?=\s*#/,
 	// Note: XSS check is done below after string is trimmed
 	rquickExpr = /^([^<]*)(<[\w\W]+>)([^>]*)$/;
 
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -3,6 +3,7 @@ var matched, browser,
 	oldInit = jQuery.fn.init,
 	oldParseJSON = jQuery.parseJSON,
 	rspaceAngle = /^\s*</,
+	rattrHash = /\[\s*\w+\s*[~|^$*]?=\s*#/,
```
gibson042: We can be even more sure that nothing invalid falls through the cracks (e.g., `"[attr=foo#bar]"`) by inserting `(?![\s'"])[^#\]]*` before the final `#`.

dmethvin: OIC, the [exception](https://github.com/jquery/sizzle/commit/f204a6112216f31685717d9fc1bf6cabf42b2ef1) was for `#` anywhere. Yeah i can do that.

dmethvin:
Fixed #140 via 2ad19820ce2d28cb47bb6d7a634707f921e7a3a1

PR #143, opened 2016-02-04
Build: Add Istanbul for code coverage reports

dmethvin:
Coverage test is running against the Phantom build but it's still useful for some basic sanity checking. It will become more so once we move to supporting only a single version of jQuery and don't conditionally fill APIs.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -65,6 +65,15 @@ module.exports = function(grunt) {
 			}
 		},
 		qunit: {
+			options: {
+				coverage: {
+					disposeCollector: true,
+					instrumentedFiles: "temp/",
+					src: [ "dist/jquery-migrate.js" ],
+					htmlReport: "coverage/",
+					linesThresholdPct: 85
+				}
+			},
 			files: [ "test/**/*.html" ]
 		},
 		npmcopy: {
@@ -122,7 +131,7 @@ module.exports = function(grunt) {
 	grunt.loadNpmTasks("grunt-contrib-watch");
 	grunt.loadNpmTasks("grunt-contrib-jshint");
 	grunt.loadNpmTasks("grunt-contrib-uglify");
-	grunt.loadNpmTasks("grunt-contrib-qunit");
+	grunt.loadNpmTasks("grunt-qunit-istanbul");
 	grunt.loadNpmTasks("grunt-npmcopy");
 
 	// Default task.
--- a/package.json
+++ b/package.json
@@ -27,15 +27,17 @@
     "grunt": "~0.4.5",
     "grunt-contrib-concat": "0.5.1",
     "grunt-contrib-jshint": "0.11.0",
-    "grunt-contrib-qunit": "0.5.2",
     "grunt-contrib-uglify": "0.8.0",
     "grunt-contrib-watch": "0.6.1",
     "grunt-git-authors": "3.0.0",
     "grunt-npmcopy": "0.1.0",
+    "grunt-qunit-istanbul": "0.6.0",
     "jquery": "^2.1.3",
-    "phantomjs-polyfill": "0.0.1",
     "qunitjs": "1.17.1",
     "testswarm": "~1.1.0"
+  },
+  "scripts": {
+    "prepublish": "grunt"
   },
   "keywords": []
 }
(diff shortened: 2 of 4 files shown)


mgol:
Wow, I wouldn't expect it to be so easy to add. :)

markelog:
That's instanbul over phantom, without need for ajax requests, shouldn't be hard at all

package.json
```diff
@@ -27,15 +27,17 @@
...
     "grunt-git-authors": "3.0.0",
     "grunt-npmcopy": "0.1.0",
+    "grunt-qunit-istanbul": "0.6.0",
     "jquery": "^2.1.3",
-    "phantomjs-polyfill": "0.0.1",
     "qunitjs": "1.17.1",
     "testswarm": "~1.1.0"
   },
+  "scripts": {
+    "prepublish": "grunt"
```
markelog: ?

markelog:
Might be good idea to add integration with coveralls too

Gruntfile.js
```diff
@@ -122,7 +131,7 @@ module.exports = function(grunt) {
 	grunt.loadNpmTasks("grunt-contrib-watch");
 	grunt.loadNpmTasks("grunt-contrib-jshint");
 	grunt.loadNpmTasks("grunt-contrib-uglify");
-	grunt.loadNpmTasks("grunt-contrib-qunit");
+	grunt.loadNpmTasks("grunt-qunit-istanbul");
```
markelog: Maybe you should consider using [load-grunt-tasks](https://github.com/sindresorhus/load-grunt-tasks) like we do in core?

package.json
```diff
@@ -27,15 +27,17 @@
     "grunt": "~0.4.5",
     "grunt-contrib-concat": "0.5.1",
     "grunt-contrib-jshint": "0.11.0",
-    "grunt-contrib-qunit": "0.5.2",
```
markelog: So you always want to generate coverage info?

package.json
```diff
@@ -27,15 +27,17 @@
...
     "grunt-contrib-uglify": "0.8.0",
     "grunt-contrib-watch": "0.6.1",
     "grunt-git-authors": "3.0.0",
     "grunt-npmcopy": "0.1.0",
+    "grunt-qunit-istanbul": "0.6.0",
     "jquery": "^2.1.3",
-    "phantomjs-polyfill": "0.0.1",
```
markelog: Why? Also you should have run `grunt npmcopy` if you removing it?

markelog:
Also, you might as well add integration with travis

markelog:
@dmethvin i see you added travis, but what about my other comments?

dmethvin:
The prepublish is gone, something added it automatically. The phantomjs-polyfill isn't needed with grunt-qunit-istanbul afict, the tests pass fine without it so it's not a direct dependency. And if i don't generate coverage info regularly i won't look at it :smile_cat:

markelog:
If you modify the package dependencies, you have to run `grunt npmcopy`, since we don't use other means.

In other words polyfill is still there - https://github.com/jquery/jquery-migrate/tree/master/external/phantomjs-polyfill

**And** - https://github.com/jquery/jquery-migrate/blob/a79bc6e9bf9e9a4aa3b2bb273db4f99735ba61dd/test/index.html#L12

But if you remove it, the tests will fail, `grunt-qunit-istanbul` doesn't add their own polyfill

> And if i don't generate coverage info regularly i won't look at it

If you integrate with coveralls, you can add a badge to the readme and also see how coverage changed with every pull request, since coveralls bot will tell you about it.

For example see express repo - https://github.com/expressjs/express

Also, what about `load-grunt-tasks` package?

markelog:
> since coveralls bot will tell you about it.

I meant hook there

dmethvin:
Oh man I forgot about npmcopy, just like I used to forget about bowercopy. The dependencies don't change often enough for it to be a habit. I can add the load-grunt-tasks stuff but in some ways I like doing it explicitly so I know what I've loaded.

markelog:
> I like doing it explicitly so I know what I've loaded.

Ok-ok, totally reasonable

What about coveralls? I can provide a pull for that this weekend, if that too big task to dig.

dmethvin:
Sure, sounds good. I'm sure you can do it in your sleep!

PR #151, opened 2016-02-25
Core: Always show a startup log message, even for minified builds

### LINKED ISSUE #149: Always console.warn or console.error when migrate is included
To deter people from including migrate in production.

"Please remove migrate in production"

dmethvin:
Fixes #149


mgol:
Should it maybe urge people more to remove Migrate in production rather than just informing them that it's active?

dmethvin:
The documentation does that but the console log message does not say that directly. I have a feeling we'll experience a big shitstorm enough from having any console message.

mgol:
I don't know, a lot of sites log various things to the console and people don't seem to care. We've received bug reports about browser *warnings* that our code triggered, I'm not sure if that many people cared if it were regular log messages not even trying to suggest something is done wrong but purely informational.

Maybe "Please remove Migrate in production" sounds a bit strong as it immediately judges... But I'm not sure if a pure "Migrate is installed" is going to ring a bell that something's wrong for most people.

dmethvin:
People who use the minified version are happy with their silence though, and anything that peeks through and nags them that they have "done something wrong" will cause them to come and express their unhappiness that they have console messages. 

I like the idea of having a message always show, because it does make things easier to debug when you wonder why a recent version of jQuery is acting bad. I just don't want to make the message too preachy, they can come look up the meaning if they are not sure what it means.

dsifford:
This PR makes me sad. Mainly because migrate is bundled with WordPress and there is no getting rid of it without running the risk of breaking plugins that depend on it. 

The result has everybody else who is developing wordpress plugins or themes having to deal with the annoying `JQMIGRATE: Migrate is installed, version 1.4.1` every time the console is opened.

dmethvin:
If your theme and plugins don't need Migrate, just dequeue it or don't enqueue it. The console reminder is really handy when you're trying to figure out performance issues.

dsifford:
@dmethvin I was under the impression that if I dequeue it for my plugin, migrate will be dequeued across the entire site (which will, in turn, affect other plugins that the user has installed that I have no control over). Is this not the case?

dsifford:
Disclosure: I really have no idea how migrate works. I just know that it exists. So I apologize if I sound ignorant!

dmethvin:
No, you are correct that if there are any plugins or components that use some older version of jQuery they may still require Migrate. Some plugins install their own, older, versions of jQuery and those may use `.noConflict()` and install new versions of Migrate. For someone trying to debug a mess like that, the Migrate console messages are again a useful tool.

dsifford:
Understood. Thanks for clarifying! 😄 

Sorry for digging this one out of the grave. Hope I didn't come across the wrong way. Thanks again so much for your work on this, and the education on the utility of Migrate.

PR #154, opened 2016-02-27
Build: enable coveralls repro & allow testing \w building

markelog:
Make you define two module list though :/

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -61,16 +61,16 @@ module.exports = function( grunt ) {
 				coverage: {
 					disposeCollector: true,
 					instrumentedFiles: "temp/",
-					src: [ "dist/jquery-migrate.js" ],
-					htmlReport: "coverage/",
-					lcovReport: "coverage/",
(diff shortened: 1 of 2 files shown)


markelog:
@dmethvin - https://coveralls.io/builds/5230460/source?filename=src%2Fmanipulation.js

dmethvin:
I like it! Too bad we need another module list though.

markelog:
At least, although not the same, we are bad everywhere - 
https://github.com/jquery/jquery/blob/8a91f8442f4d31bdd4b43035ee88d40fbb116f64/Gruntfile.js#L132
https://github.com/jquery/jquery/blob/8a91f8442f4d31bdd4b43035ee88d40fbb116f64/test/data/testinit.js#L279

dmethvin:
Yeah I always forget to change it in every place whenever I'm adding a module. :smiley_cat:

mgol:
If only we had a cross-environment module system...

Issue #87, opened 2015-02-09
Dashes in .data() names behavior change


dmethvin:
We're using camelCase for data names where possible, and only use the dash-name if it exists in the object already. Ideally, API callers should use camelCase for accessing data names rather than dashes. If the caller gets the raw data object they'll see camelCase if they put in dash-name via the API. (Clearly this will need a longer explanation somewhere.)

https://github.com/jquery/jquery/issues/1751
https://github.com/jquery/jquery/issues/2257

dmethvin:
API docs need some variant of this description: https://github.com/jquery/jquery/issues/1751#issuecomment-67721150

dmethvin:
At this point I'm not sure how to make a non-noisy warning about this. It might just have to be a documentation change.

gibson042:
I think this can be done: http://jsfiddle.net/aq992xxb/

The following cases will need warnings:
* Dashed key in a bulk setter (`$el.data({ "dashed-key": … })` does not camelCase before 3.0)
* Dashed key getter when an exact match key exists in the data object (`$el.data()["dashed-key"] = …; $el.data("dashed-key")` retrieves the exact match instead of the camelCase before 3.0)
* Update dashed key when an exact match key exists in the data object (`$el.data()["dashed-key"] = …; $el.data("dashed-key", newValue)` updates both values instead of just the camelCased one in 2.x)

dmethvin:
We're only talking about breaking changes with regard to migration though. So we only want to give a warning if some old code no longer works. Merely using dashed keys needs to be silent unless mixed with direct data object access where the new representation would be visible?

gibson042:
Pretty much, yeah. Access is backwards- incompatible only when the data object includes a dashed key, and _that_ can only happen through a bulk setter call (which itself is one of those warning cases) or direct manipulation.

dmethvin:
@gibson042 How does this sound? In a call to `.data("some-name")` the 3.0 behavior is to only look for `someName` in the data item, even if `some-name` is set. Migrate can do the probe for `some-name` and see if it is present in the data object, then return that value instead of `someName` and warn about it. That's the only case I see that's easy to detect.

PR #155, opened 2016-03-03
Core: Remove workarounds for APIs dropped before 3.0

### LINKED ISSUE #152: Remove the code from jQuery versions older than 1.12/2.2

mgol:
Fixes #152

I tried to pay attention to what I'd been removing but I'd apprieciate a 2nd (or 3rd) pair of eyes. :D

The build is failing because of low test coverage (79.65%, below the 85% threshold) as I mainly removed well tested code. I'll add some tests once it's verified I haven't removed too much.


mgol:
The module is not really `Core` but lots of modules but that wouldn't even fit in our message limit.

mgol:
Ah, coverage is fine, I just haven't run `grunt build` before running `npm test`. We should get the npm scripts to be equivalent to what we have in Core, current setup is confusing.

markelog:
Nice!

> I just haven't run grunt build before running npm test.

Hm, `npm test` should work without build, what are repro steps?

mgol:
Steps:
1. Run `grunt` on `master`.
1. Apply my PR.
1. Run `npm test`.

markelog:
Can't repro -
```sh
$ grunt
Running "concat:dist" (concat) task
File dist/jquery-migrate.js created.

Running "uglify:all" (uglify) task
>> 1 file created.

Running "jshint:dist" (jshint) task
>> 1 file lint free.

Running "jshint:tests" (jshint) task
>> 12 files lint free.

Running "jshint:grunt" (jshint) task
>> 1 file lint free.

Running "jscs:src" (jscs) task
>> 27 files without code style errors.

Running "qunit:files" (qunit) task
Testing test/index.html JQMIGRATE: Migrate is installed with logging active, version 1.3.1-pre
...............................................OK
>> 774 assertions passed (1637ms)
>> Coverage:
>> -  Lines: 86.04%
>> -  Statements: 86.04%
>> -  Functions: 90%
>> -  Branches: 70.33%

Done, without errors.
$ g fpr 155
From github.com:jquery/jquery-migrate
 * [new ref]         refs/pull/155/head -> 155
Switched to branch '155'
$ npm test

> jquery-migrate@3.0.0-pre test /Users/arkel/Workspace/jquery-migrate
> grunt test

Running "qunit:files" (qunit) task
Testing test/index.html JQMIGRATE: Migrate is installed with logging active, version 1.3.1-pre
.....................OK
>> 62 assertions passed (1211ms)
>> Coverage:
>> -  Lines: 87.25%
>> -  Statements: 87.25%
>> -  Functions: 88.89%
>> -  Branches: 70.49%

Done, without errors.
```

Gruntfile.js
```diff
@@ -21,11 +21,8 @@ module.exports = function( grunt ) {
 			"src/intro.js",
 			"src/version.js",
 			"src/migrate.js",
-			"src/attributes.js",
 			"src/core.js",
 			"src/css.js",
-			"src/data.js",
```
dmethvin: This may be back soon if #87 is fixed, but we might as well take it out now.

src/migrate.js
```diff
@@ -39,36 +39,6 @@ function migrateWarn( msg ) {
 	}
 }
 
-function migrateWarnProp( obj, prop, value, msg ) {
```
dmethvin: Wow, we don't need this anymore? It's another one I suspect will be back, but might as well remove it. If it came back we wouldn't need the guard for `Object.defineProperty` or the `try/catch` anyway.

src/core.js
```diff
@@ -1,52 +1,16 @@
...
-			// JQuery object as context; parseHTML expects a DOM object
-			context = context.context;
-		}
-		if ( jQuery.parseHTML ) {
-			return oldInit.call( this,
-					jQuery.parseHTML( match[ 2 ], context && context.ownerDocument ||
-						context || document, true ), context, rootjQuery );
-		}
-	}
+// JQuery( "#" ) is a bogus ID selector, but it returned an empty set before jQuery 3.0
```
dmethvin: This comment doesn't seem needed since it's duplicated in the body.

test/core.js
```diff
@@ -105,104 +69,64 @@ test( "attribute selectors with naked '#'", function() {
...
 
-	test = jQuery( document.body ).find( "#qunit-fixture" );
-	equal( test.selector, "#qunit-fixture", "#qunit-fixture find Selector" );
-	equal( test.context, document.body, "#qunit-fixture find Context" );
+	equal( threw, true, "Throw on leading-hash HTML (treated as selector)" );
 } );
 
-test( "XSS injection", function() {
-	expect( 10 );
+test( "XSS injection (XSS via script tag)", function() {
```
dmethvin: It doesn't hurt to keep these, but since we [pulled out](https://github.com/jquery/jquery-migrate/pull/155/files#diff-7eb52b366866677666470e019283c8eaL25) the check above we shouldn't need it should we?

mgol:
@markelog OK, the steps I provided don't work but it definitely happened to me. I was working on the patch, removing more & more stuff & running `npm test` in the meantime (which I sometimes forgot to preceed by `grunt build`) and that's what I got:
```
npm test

> jquery-migrate@3.0.0-pre test /Users/mgol/Documents/projects/public/jquery/jquery-migrate
> grunt test

Running "qunit:files" (qunit) task
Testing test/index.html JQMIGRATE: Migrate is installed with logging active, version 1.3.1-pre
.....................OK
>> 62 assertions passed (1187ms)
>> Coverage:
>> -  Lines: 79.65%
>> -  Statements: 79.65%
>> -  Functions: 76.19%
>> -  Branches: 68.25%
Warning: Code coverage for lines was below threshold (79.65% < 85%) Use --force to continue.

Aborted due to warnings.
npm ERR! Test failed.  See above for more details.
```

Even regardless of this issue I'd still want to see the `build` script to be prepended by `npm install` and the `test` one by `grunt build` as in https://github.com/jquery/jquery/blob/93a8fa6bfc1c8a469e188630b61e736dfb69e128/package.json#L59-L64. Let's be consistent.

test/core.js
mgol: I thought about it but I thought it might be worth to keep them as long as we patch `jQuery.fn.init` in any way, just in case. No strong opinions here, though.

src/core.js
mgol: Fixed.

src/migrate.js
mgol: Yeah, JSHint yelled at me because it was unused. :)

Gruntfile.js
mgol: I may leave it empty if you wish but adding new modules isn't that hard anyway. :)

markelog:
@mgol Sounds like you were on intermediate step - removed the tests, but not the code, or didn't applied latest migrate patches i.e. before the rebase.

Difference between the migrate and core, that in the latter you have to build (if not running in the amd mode which not default) before run the tests, which is not necessary for the migrate, since migrate, by default, would still not use your builded version.

mgol:
> Difference between the migrate and core, that in the latter you have to build (if not running in the amd mode which not default) before run the tests, which is not necessary for the migrate, since migrate, by default, would still not use your builded version.

Even then, `npm test` should invoke all tests, including running linters and since both Core & Migrate have additional JSHint setup for the built file, this require building. I'd like to align with Core here.

markelog:
Mmm, okay, sounds logical, when you think about that, a lot of projects do the same like in `babel-register` case for example, plus alignment would also be nice.

markelog:
If do change that though, make sure it would run in `min` mode, since now it runs in `dev` one, but that should happen only at `npm t` time, otherwise linting the min source without running the tests in one, is not coherent. But it also would be nice, to keep running the suite, without needing to build it, that is why i have "only at `npm t` time" note.

dmethvin:
Changes LGTM here.

mgol:
Landed. 770 assertions before, 62 after. The new Migrate will be *really* new. ;)

mgol:
I created #157 for the npm scripts refactor plan.

PR #158, opened 2016-03-23
Build: Only support jQuery 3.0 and master

dmethvin:
@mgol Does this look ok to you?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,11 +31,10 @@ module.exports = function( grunt ) {
 		],
 		tests: {
 			"jquery": [
-				"dev+2.x-git",
-				"min+2.x-git.min",
-				"dev+2.2.2",
-				"dev+1.x-git",
-				"dev+1.12.2"
+				"dev+git",
+				"min+git.min",
+				"dev+3.0.0",
+				"dev+3.0.0.min"
 			]
 		},
 		banners: {
--- a/test/index.html
+++ b/test/index.html
@@ -3,8 +3,6 @@
 <head>
 	<meta charset="utf-8">
 	<title>jQuery Migrate Test Suite</title>
-	<!-- Ensure IE doesn't fall back into oldIE modes -->
-	<meta http-equiv="X-UA-Compatible" content="IE=Edge" />
 
 	<!-- QUnit -->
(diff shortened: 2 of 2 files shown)


test/index.html
```diff
@@ -3,8 +3,6 @@
 <head>
 	<meta charset="utf-8">
 	<title>jQuery Migrate Test Suite</title>
-	<!-- Ensure IE doesn't fall back into oldIE modes -->
-	<meta http-equiv="X-UA-Compatible" content="IE=Edge" />
```
mgol: Why did you remove this? we still test on IE9-11 so it seems it's better to leave it there.

Gruntfile.js
```diff
@@ -31,11 +31,10 @@ module.exports = function( grunt ) {
...
 		tests: {
 			"jquery": [
-				"dev+2.x-git",
-				"min+2.x-git.min",
-				"dev+2.2.2",
-				"dev+1.x-git",
-				"dev+1.12.2"
+				"dev+git",
+				"min+git.min",
+				"dev+3.0.0",
```
mgol: This will work only after the final 3.0.0 release, won't it? We should probably put `3.0.0-beta` for now. Or wait with this entry for a final release and for now just do the `git` part.

test/index.html
dmethvin: In IE8 there used to be a button next to the URL to switch into IE7 Compatibility View, which people accidentally pressed. I thought they removed it in IE9, but I just checked and it didn't go away until IE10. Since IE10 the only way for any IE10+ browser to be forced into Compat View by the user is to either manually add it in a deeply buried dialog or use Enterprise Mode and have their IT department list it in a config file. That's super unlikely in our unit test scenario.

Bottom line is that you're right, since it *is* possible to push that button in IE9 I'll restore it.

Gruntfile.js
dmethvin: Yeah, since I didn't care about the beta I figured that testing with `git` was good enough for now and I was just getting a running start on the release. :smiley_cat:

test/index.html
mgol: I think the button only went away in IE 11, not 10.

PR #160, opened 2016-03-25
Offset: Fill and warn if .offset() is called with strange inputs

### LINKED ISSUE #108: Restore .offset disconnected/invalid behavior
As of jQuery 3.0, disconnected or non-element input to `.offset` will no longer return `{ top: 0, left: 0 }`, and may throw. Migrate should restore functionality changed by https://github.com/jquery/jquery/commit/1617479fcf7cbdaf33dc9334ed10a0f30bf14687 / https://github.com/jquery/jquery/commit/d4dd548acaf049d8e4dca9e4b879a51f26bb3d27, and warn when there is no layout box.

dmethvin:
Fixes #108

### DIFF
--- a/test/offset.js
+++ b/test/offset.js
@@ -1,0 +1,30 @@
+
+module( "offset" );
+
+QUnit.test( ".offset()", function( assert ) {
+	expect( 8 );
+
+	var bogus = { top: 0, left: 0 };
+
+	expectNoWarning( ".offset() on body", function() {
+		assert.ok( "top" in jQuery( "body" ).offset(), "got an offset object" );
+	} );
+
+	expectWarning( ".offset() on window", function() {
+		assert.deepEqual(
+			jQuery( window ).offset(),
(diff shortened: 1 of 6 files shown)


test/offset.js
```diff
@@ -0,0 +1,30 @@
...
+			jQuery( document.createElement( "div" ) ).offset(),
+			bogus, "disconnected bogus top/left 0" );
+	} );
+
+	expectWarning( ".offset() on plain object", function() {
+		assert.deepEqual(
+			jQuery( { space: "junk", zero: 0 } ).offset(),
+			bogus, "plain object bogus top/left 0" );
+	} );
+} );
```
mgol: One example that ensures the regular `.offset()` usage is not broken would be good to have.

dmethvin: The first was intended to do that but I now realize it doesn't do that. Plus I forgot to check that `.offset()` as a setter still works. Thanks!

PR #162, opened 2016-03-25
Ajax: Warn and fill the jQXHR .success, .error, and .complete methods

### LINKED ISSUE #88: Deprecated Ajax methods removed.
https://github.com/jquery/jquery/issues/2084

`jqXHR.success | error | complete`

dmethvin:
Fixes #88 

This reintroduces a streamlined `migrateWarnProp` since we don't need to support a missing broken `Object.defineProperty` like IE8. That also means it will issue the warning on property access rather than wait until the method call, but that won't matter in most cases. In the cases where it does matter, you probably want to know where the property was accessed.

### DIFF
--- a/src/migrate.js
+++ b/src/migrate.js
@@ -46,10 +46,6 @@ function migrateWarnProp( obj, prop, value, msg ) {
 		get: function() {
 			migrateWarn( msg );
 			return value;
-		},
-		set: function( newValue ) {
-			migrateWarn( msg );
-			value = newValue;
 		}
 	} );
 }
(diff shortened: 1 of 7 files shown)


src/migrate.js
```diff
@@ -39,6 +39,21 @@ function migrateWarn( msg ) {
...
 
+function migrateWarnProp( obj, prop, value, msg ) {
+	Object.defineProperty( obj, prop, {
+		configurable: true,
+		enumerable: true,
+		get: function() {
+			migrateWarn( msg );
+			return value;
+		},
+		set: function( newValue ) {
```
dmethvin: As coveragebot observed, this is not being checked. I'll add a test.

PR #164, opened 2016-03-29
Core: Ensure jQuery 3.0+ and only one Migrate is loaded

### LINKED ISSUE #115: Support only the 3.x release of jQuery Core
As of jQuery Migrate 3.0, this plugin should be paired only with Core 3.x and not older versions. That simplifies the work the plugin needs to do since it can depend on the behavior and APIs of the latest core and not have to work around all the quirks of previous versions. 

* [ ] Document this new usage in the README, which is to include the latest Core and Migrate in the old project. (Previously, we would tell users to include Migrate with the current version of Core they were using on the page.)
* [ ] Add a version check to ensure people don't misuse 3.0 by running it with older versions of jQuery.
* [x] Update Migrate code to remove any Core-version-dependent code paths, for example the [checks whether internal functions are privatized](https://github.com/jquery/jquery-migrate/blob/master/src/css.js#L4)
* [x] Determine and document the Core versions that we support for doing an upgrade. Presently it's 1.6.4 but this approach may make it possible to go back further?

dmethvin:
Fixes #115 

Coverage will go down a bit because it's testing for situations that can't easily be checked in unit tests. I've checked them locally. It could be done by pulling over something like testIframe from core but I'm okay with light testing on this right now since I don't want it to hold up the release.


README.md
```diff
@@ -1,33 +1,34 @@
...
 
 All warnings generated by this plugin start with the string "JQMIGRATE". A list of the warnings you may see are in [warnings.md](https://github.com/jquery/jquery-migrate/blob/master/warnings.md).
 
 ### Development version
 
-This version provides console warning messages when deprecated and/or removed APIs are used. Use this version during development and debugging, and whenever you are reporting bugs to the jQuery team.
+The development version provides console warning messages when deprecated and/or removed APIs are used. Use this version during development and debugging, and whenever you are reporting bugs to the jQuery team.
 
 **Latest released development version:** This file is hosted on jQuery's CDN, and can be hotlinked if desired.
 [http://code.jquery.com/jquery-migrate-1.3.0.js](http://code.jquery.com/jquery-migrate-1.3.0.js)
```
dmethvin: These references to 1.3 will be updated by the release script on our first release.

README.md
```diff
@@ -1,33 +1,34 @@
 [![Build Status](https://travis-ci.org/jquery/jquery-migrate.svg?branch=master)](https://travis-ci.org/jquery/jquery-migrate)
 [![Coverage Status](https://img.shields.io/coveralls/jquery/jquery-migrate.svg?style=flat)](https://coveralls.io/r/jquery/jquery-migrate?branch=master)
 
-# Migrate older jQuery code to jQuery 1.9+
+# Migrate older jQuery code to jQuery 3.0+
 
-This plugin can be used to detect and restore APIs or features that have been deprecated in jQuery and removed as of version 1.9.
+#### NOTE: If you have not yet upgraded to 1.12.x or 2.2.x, first upgrade to them and use [jQuery Migrate 1.x](https://github.com/jquery/jquery-migrate/tree/1.x-stable#readme) to resolve any compatibility issues before using jQuery Migrate 3.0 and upgrading to jQuery 3.0.
+
+This plugin can be used to detect and/or restore APIs and behaviors that have been deprecated and/or removed as of jQuery version 3.0.
```
gibson042: "detect and/or restore" doesn't read quite right... I think it should be "warn about and/or restore".

src/migrate.js
```diff
@@ -1,18 +1,34 @@
...
+		rbadVersions = /^[12]\./;
 
-// Set to true to prevent console output; migrateWarnings still maintained
-// jQuery.migrateMute = false;
+	if ( !log ) {
+		return;
+	}
+
+	// Need jQuery 3.0.0+ and no older Migrate loaded
+	if ( !window.jQuery || rbadVersions.test( jQuery.fn.jquery ) ) {
```
gibson042: Drop the `window.` prefix; `jQuery` is already a [local variable](https://github.com/dmethvin/jquery-migrate/blob/115-only3/src/intro.js).

test/old-jquery.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<title>jQuery Migrate old-jQuery initialization test</title>
+	<!-- // Support: IE < 11 -->
```
mgol: This should say `Support: IE <=10 only`. ;-)

PR #166, opened 2016-03-29
Event: Add triggering for "ready" event

### LINKED ISSUE #136: Add back ready event
Ready event is removed in 3.0

https://github.com/jquery/jquery/issues/2264

dmethvin:
Fixes #136 

Another one that would seem to need an iframe to test properly.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,7 +63,7 @@ module.exports = function( grunt ) {
 					linesThresholdPct: 85
 				}
 			},
-			files: [ "test/**/*.html" ]
+			files: [ "test/**/index.html" ]
 		},
 		coveralls: {
 			src: "coverage/lcov/lcov.info",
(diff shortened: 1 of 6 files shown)


dmethvin:
Just to clarify what's going on with this PR, we already give a warning when a `"ready"` event is attached to `document` and there are unit tests verifying that. What we are missing is only the code to trigger the `"ready"` event, since before 3.0 it was triggered by core. Since the event only fires once on document ready, we would probably need to set up an iframe in a separate document to catch the event firing at the right time.

mgol:
Shouldn't we do iframe tests like Core does?

dmethvin:
Yeah, I'm working on it.

dmethvin:
This is failing because of https://github.com/jquery/jquery/issues/3040 due to https://github.com/jquery/jquery/commit/5cbb234dd3273d8e0bbd454fb431ad639c7242c1, I was trying to figure out why it was working fine over the weekend. :smile_cat:

PR #169, opened 2016-04-08
Core: Warn about jQuery.isNumeric differences

### LINKED ISSUE #134: Restore 2.2 isNumeric signature
jquery/jquery/issues/2662

jQuery 3 changed `jQuery.isNumeric` to only return true on number/string input, whereas it would previously use `toString` and exempt only arrays. Migrate should restore and warn about the old behavior.

dmethvin:
Closes #134 

It just seemed simplest to put the old implementation in the code and compare the two.

Are there any other edge cases we need to test here? Is there a better warning message?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,5 +1,6 @@
 
 var oldInit = jQuery.fn.init,
+	oldIsNumeric = jQuery.isNumeric,
 	rattrHash = /\[\s*\w+\s*[~|^$*]?=\s*(?![\s'"])[^#\]]*#/;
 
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -1,5 +1,6 @@
 
 var oldInit = jQuery.fn.init,
+	oldIsNumeric = jQuery.isNumeric,
```
dmethvin: Dunno why I did this, I could have called it directly since I'm not patching it. :smile_cat:

gibson042: Not at all; LGTM.

PR #171, opened 2016-04-11
Update jQuery Migrate links to v1.4.0

coliff:
Update jQuery Migrate links to v1.4.0


dmethvin:
Okay, I see what the problem was. I hadn't landed the README update that was in #164, which I just did. The README is for the 3.0 version of Migrate and has a pointer to the 1.x branch which has the correct 1.4.0 version. Take a look, does that seem correct now?

coliff:
Ah, I see. Makes sense now - looks good! :+1:

PR #172, opened 2016-04-12
Attributes: Warn and fill removeAttr boolean properties

### LINKED ISSUE #118: removeAttr( boolean ) no longer sets the property to false
https://github.com/jquery/jquery/issues/1759

dmethvin:
Fixes #118


src/attributes.js
```diff
@@ -0,0 +1,16 @@
+
+var oldRemoveAttr = jQuery.fn.removeAttr,
+	rmatchNonSpace = /\S+/g;
+
+jQuery.fn.removeAttr = function( name ) {
+	var self = this;
+
+	jQuery.each( name.match( rmatchNonSpace ), function( i, attr ) {
+		if ( jQuery.expr.match.bool.test( attr ) ) {
```
dmethvin: @timmywil @gibson042 How do you feel about using `jQuery.expr.match.bool` here?

timmywil: Seems right.

PR #176, opened 2016-04-22
fix(README): point to v1.4.0 release urls

thgreasi:


dmethvin:
This isn't the latest README on master. Take a look at that one. Is the information you need there?

thgreasi:
I'm a bit puzzled by your answer.
I think that I'm targeting the [README of the `jquery:master` branch](https://github.com/jquery/jquery-migrate/blob/master/README.md) which only includes reference to v1.3.0.
Do you mean the file in the (1.x-stable branch)[https://github.com/jquery/jquery-migrate/blob/1.x-stable/README.md] (which I just realized that existed)?

So, I guess you can close this, if the `master` branch is only used for jquery v3.0 development...
On the other hand, others might get confused (like I did) when landing on the main page of the repo.

dmethvin:
Oh okay I see what you mean. Those will be updated to the 3.0 branch as soon as the first 3.0 release is made, the build script does it. If you're looking for 1.x code then the link up top should point to the README for that branch, and it has accurate information for the 1.x branch since a release has already been run against it. I can land this for now though, but people looking for the 1.x code should not be reading further into the README here because it doesn't apply to the 1.x branch.

dmethvin:
Landed at ee0a030bfc95e25265a46c072cd1186db85d1872

Thanks! Had you noticed the big link to 1.x before? I think this might become more of an issue when the first 3.0 release ships if a lot of people are looking for 1.x and get 3.x by accident, but I tried to make the link at the top as large as possible.

thgreasi:
Interestingly enough I did notice the v3.0 header but it looks like my
brain just skipped that and searched for a cdn url. :+1:

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

Issue #179, opened 2016-04-27
switch to "use strict"


dmethvin:
Since core is now using strict, Migrate should as well.

mgol:
In Migrate it should be simpler as there are no AMD modules so it's all the question of adding one `"use strict";` in intro. :)

dmethvin:
Yeah and it seems to be working locally at least.

PR #180, opened 2016-04-27
Build: Add code for strict mode

dmethvin:
Not sure if I should have any other tests here. I don't need or need to put all the tests into strict mode.

### DIFF
--- a/src/intro.js
+++ b/src/intro.js
@@ -1,2 +1,2 @@
-(function( jQuery, window, undefined ) {
+(function( jQuery, window ) {
 "use strict";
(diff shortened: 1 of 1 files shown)


dmethvin:
Duh fixes #179

mgol:
Interesting, no changes required to pass the tests? :)

LGTM then!

src/intro.js
```diff
@@ -1,3 +1,2 @@
 (function( jQuery, window, undefined ) {
```
mgol: BTW, we could drop the `undefined`; Core doesn't protect itself from `undefined` shadowing anymore so there's no reason Migrate should.

dmethvin:
I tried it on all the desktop browsers and Android 4.1, they seemed to be fine. I even tried it on IE8 not really thinking. It doesn't work there. :stuck_out_tongue: We'll see if the whole suite has any issues.

PR #181, opened 2016-04-28
Attributes: Warn and fill .toggleClass( boolean )

### LINKED ISSUE #175: Warn and fill `.toggleClass()` and `.toggleClass( boolean )`
https://github.com/jquery/jquery/issues/2491

https://github.com/jquery/jquery/commit/a4d16a26abe3ea36337f7bfc4d1bfe13b893cd17

dmethvin:
Fixes #175

### DIFF
--- a/test/attributes.js
+++ b/test/attributes.js
@@ -41,3 +41,40 @@ QUnit.test( ".removeAttr( boolean attribute )", function( assert ) {
 	} );
 
 } );
+
+QUnit.test( "", function( assert ) {
+	assert.expect( 13 );
+
+	var e = jQuery( "<div />" ).appendTo( "#qunit-fixture" );
+
+	expectWarning( "toggling initially empty class", function() {
+		e.toggleClass( true );
+		assert.equal( e[ 0 ].className, "", "Assert class is empty (data was empty)" );
+	} );
+
+	expectNoWarning( "toggleClass not full className", function() {
+		e.attr( "class", "" );
+		e.toggleClass( "classy" );
+		assert.equal( e.attr( "class" ), "classy", "class was toggle-set" );
+		e.toggleClass( "classy", false );
+		assert.equal( e.attr( "class" ), "", "class was toggle-removed" );
+	} );
+
+	expectWarning( "toggleClass full-className manipulations", function() {
+		e.addClass( "testD testE" );
+		assert.ok( e.is( ".testD.testE" ), "Assert class present" );
+		e.toggleClass();
+		assert.ok( !e.is( ".testD.testE" ), "Assert class not present" );
+		assert.equal( jQuery.data( e[ 0 ], "__className__" ), "testD testE", "Data was stored" );
+		e.toggleClass();
+		assert.ok( e.is( ".testD.testE" ), "Assert class present (restored from data)" );
+		e.toggleClass( false );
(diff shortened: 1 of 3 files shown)


dmethvin:
Note to self: Need to update the upgrade guide, it currently says only the no-args form is deprecated but the single-boolean form is deprecated as well.

test/attributes.js
```diff
@@ -41,3 +41,40 @@ QUnit.test( ".removeAttr( boolean attribute )", function( assert ) {
...
+		assert.equal( e.attr( "class" ), "classy", "class was toggle-set" );
+		e.toggleClass( "classy", false );
+		assert.equal( e.attr( "class" ), "", "class was toggle-removed" );
+	} );
+
+	expectWarning( "toggleClass full-className manipulations", function() {
+		e.addClass( "testD testE" );
+		assert.ok( e.is( ".testD.testE" ), "Assert class present" );
+		e.toggleClass();
+		assert.ok( !e.is( ".testD.testE" ), "Assert class not present" );
```
mgol: Maybe test that `e` has none of those classes and not that it just doesn't have both at the same time?

test/attributes.js
```diff
@@ -41,3 +41,40 @@ QUnit.test( ".removeAttr( boolean attribute )", function( assert ) {
...
+
+	expectNoWarning( "toggleClass not full className", function() {
+		e.attr( "class", "" );
+		e.toggleClass( "classy" );
+		assert.equal( e.attr( "class" ), "classy", "class was toggle-set" );
+		e.toggleClass( "classy", false );
+		assert.equal( e.attr( "class" ), "", "class was toggle-removed" );
+	} );
+
+	expectWarning( "toggleClass full-className manipulations", function() {
```
mgol: If only the single boolean arg or no arg case was issuing errors this test would still pass. Maybe assert the number of warnings as well?

warnings.md
```diff
@@ -166,3 +166,9 @@ See jQuery-ui [commit](https://github.com/jquery/jquery-ui/commit/c0093b599fcd58
 **Cause:** The standard way to add new custom selectors through jQuery is `jQuery.expr.pseudos`. These two other aliases are deprecated, although they still work as of jQuery 3.0.
 
 **Solution:** Rename any of the older usage to `jQuery.expr.pseudos`. The functionality is identical.
+
+### JQMIGRATE: jQuery.fn.toggleClass( boolean ) is deprecated
```
mgol: `jQuery.fn.toggleClass( boolean )` and `jQuery.fn.toggleClass()`

test/attributes.js
dmethvin: These tests were copied from core pretty much as-is so I hesitate to add more assertions. Whatever it was doing before was lightly documented and we're deprecating it now.

test/attributes.js
dmethvin: Yeah I'll tighten that up.

warnings.md
dmethvin: How about `[ boolean ]` to keep the error message short? The longer description already covers both cases.

mgol: 👍

test/attributes.js
mgol: OK then!

Issue #178, opened 2016-04-25
jquery wrap() function triggers error “document.createDocumentFragment is not a function” on dynamic content load (when using migrate)


ekonoval:
Recently I've encountered a very strange behavior of wrap() method of jquery. I have a banner system and another site which uses that system. I try to update jquery version to the latest available one of the 1x branch. But I use some old plugins and would like them working with migrate fallback.

Spending a lot of hours I defined that dynamic content loading by document.write affects the work of wrap() which triggers very strange error

> TypeError: document.createDocumentFragment is not a function safeFrag = document.createDocumentFragment();

I had a few experiments with isolated code and defined that **wrap() works fine when I don't use migrate fallback** or don't use content loading by document.write.

I have no ideas why it's working this way, maybe somebody can help me? I set up a few examples

1) Expected to see "WRAPPING DONE" https://jsfiddle.net/quadroval/w0czwvbj/

`$(function() {
  $('#tryToWrap').wrap("<p>WRAPPING DONE</p>")
});`
2) Static content - wrapping works fine https://jsfiddle.net/quadroval/jLfygsgr/

3) Migration is turned off - wrapping works fine too. The content is loaded dynamically https://jsfiddle.net/quadroval/x89y0ka2/

Here you can see clear example http://sunsey.0fees.net/exper/dynamic.html

dmethvin:
The problem is the setting of `document.context=\'fHA6Mw==\';` in the code which triggers some logic inside Migrate. We can work around that but I am wondering why it is set?

Also, the `.wrap()` call isn't valid, it's trying to wrap a paragraph around a div. Paragraphs can't contain block-level elements. That's not causing the error though.

ekonoval:
well I use some old banner system openx v2.8.8 , and it generates the code like I put to jsfiddle. I also have some plugin CoinSlider [js](http://internet-bilet.ua/-js/coin-slider.js)  which uses some kind of wrap, which fails. I just simplified the code to see what exactly is causing the trouble. Strange thing that without migrate everything works fine. So I thought it may be related to migration issue. Before that I've been using jquery 1.8 and everything worked fine too.

ekonoval:
the code of coinSlider where error occurs http://prntscr.com/axifz0

> 		init = function(el){
				
			order[el.id] 		= new Array();	// order of square appereance
			images[el.id]		= new Array();
			links[el.id]		= new Array();
			linksTarget[el.id]	= new Array();
			titles[el.id]		= new Array();
[...]

PR #182, opened 2016-04-28
Core: Only check selector when it's a string

dmethvin:
Ref #178 which is a similar problem happening in 1.4, as well as a problem with a variable being set on `document`.

The solution for the 1.x branch is a bit more complicated but I figured I'd fix 3.0 first.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -28,7 +28,7 @@ test( "jQuery( '#' )", function() {
 } );
 
 test( "attribute selectors with naked '#'", function() {
-	expect( 6 );
+	expect( 7 );
 
 	// These are wrapped in try/catch because they throw on jQuery 1.12.0+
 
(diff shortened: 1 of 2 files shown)


test/core.js
```diff
@@ -67,6 +67,16 @@ test( "attribute selectors with naked '#'", function() {
 			jQuery( "a[href=\"#junk\"]" );
 		} catch ( e ) {}
 	} );
+
+	expectNoWarning( "ready function with attribute selector", function() {
+		try {
+			jQuery( function() {
+				if ( jQuery.thisIsNeverTrue ) {
+					jQuery( "a[href=#]" );
```
mgol: Since the code never reaches this statement, what's its purpose? I don't understand what's exactly tested here.

dmethvin: Because the is-it-a-string check was missing, the old code takes the function and stringifies the body during `rattrHash.test( selector )`. So if the stringified function had a pattern that looks like a bad one, it would try to give a warning when it should not--it should only give the warning when the code with the selector actually runs and we shouldn't even be checking functions here. Since the document is ready the function *will* run immediately and the `if` just  makes sure the selector isn't actually executed.

So yeah. :smile: I'm really surprised this didn't cause trouble before but I guess JavaScript will pretty much stringify anything.

mgol: Ah, I didn't think about the function stringification. So evil! :D

PR #183, opened 2016-04-28
Core: Make jQuery.fn.init more resilient

### LINKED ISSUE #178: jquery wrap() function triggers error “document.createDocumentFragment is not a function” on dynamic content load (when using migrate)

dmethvin:
Fixes #178

- Better sniff for a jQuery object, avoids document.context trap
- Don't do string checks on non-string arguments to init

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -11,45 +11,50 @@ var matched, browser,
 jQuery.fn.init = function( selector, context, rootjQuery ) {
 	var match, ret;
 
-	if ( selector && typeof selector === "string" && !jQuery.isPlainObject( context ) &&
-			(match = rquickExpr.exec( jQuery.trim( selector ) )) && match[ 0 ] ) {
-		// This is an HTML string according to the "old" rules; is it still?
-		if ( !rspaceAngle.test( selector ) ) {
-			migrateWarn("$(html) HTML strings must start with '<' character");
-		}
-		if ( match[ 3 ] ) {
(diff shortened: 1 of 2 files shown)


dmethvin:
Fixed in 1.x-stable at c09b6a01435e4b5c16e12f2476ea8456e4a7985f

PR #184, opened 2016-04-28
Attributes: Fix unquoted hashes in selectors where possible

### LINKED ISSUE #174: Try to fix the [href=something#hashed] case
https://github.com/jquery/jquery/issues/2824#issuecomment-211424520

dmethvin:
Fixes #174 

Still a WIP, I need to contemplate the regex and make the tests a little more better about checking whether the selector works. However I thought it might be good to see if anyone had concerns about this approach.

I left cases alone where I might possibly be breaking the selector even more or silently changing its meaning. However, the current approach does seem to fix all the reported cases in core tickets so far, so it's got that going for it.

A more aggressive approach would be to attempt a change even if the resulting selector isn't `qSA`-acceptable. That covers cases like `a[href=#]:first` that the current code does not. Thoughts?

Given how much this is reported I think it would be good to leave in Migrate 3.0 but also cherry-pick to 1.x.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,7 +1,7 @@
 
 var oldInit = jQuery.fn.init,
 	oldIsNumeric = jQuery.isNumeric,
-	rattrHash = /\[\s*\w+\s*[~|^$*]?=\s*(?![\s'"])[^#\]]*#/;
+	rattrHash = /\[(\s*[-\w]+\s*)([~|^$*]?=)\s*([-\w#]+)\s*\]/g;
 
 jQuery.fn.init = function( selector ) {
 	var args = Array.prototype.slice.call( arguments );
@@ -13,11 +13,30 @@ jQuery.fn.init = function( selector ) {
 			migrateWarn( "jQuery( '#' ) is not a valid selector" );
 			args[ 0 ] = selector = [];
 
-		} else if ( rattrHash.test( selector ) ) {
+		// Can't use .test here because rattrHash is //g and in global scope
+		} else if ( rattrHash.exec( selector ) ) {
 
 			// The nonstandard and undocumented unquoted-hash was removed in jQuery 1.12.0
-			// Note that this doesn't actually fix the selector due to potential false positives
-			migrateWarn( "Attribute selectors with '#' must be quoted: '" + selector + "'" );
+			// First see if qS thinks it's a valid selector, if so avoid a false positive
+			try {
+				document.querySelector( selector );
+			} catch ( err1 ) {
+
+				// Didn't *look* valid to qSA, warn and try quoting what we think is the value
+				selector = selector.replace( rattrHash, function( _, attr, op, value ) {
+					return "[" + attr + op + "\"" + value + "\"]";
+				} );
+
+				// If the regexp *may* have created an invalid selector, don't update it
+				// Note that there may be false alarms if selector uses jQuery extensions
+				try {
+					document.querySelector( selector );
+					args[ 0 ] = selector;
+					migrateWarn( "Attribute selector with '#' must be quoted: " + selector );
+				} catch ( err2 ) {
+					migrateWarn( "Attribute selector with '#' was not fixed: " + args[ 0 ] );
+				}
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -12,11 +12,31 @@ jQuery.fn.init = function( selector ) {
...
+
+			selector = selector.replace( rattrHash, function( _, attr, op, value ) {
+				return "[" + attr + op + "\"" + value + "\"]";
+			} );
+
+			// If the regexp *may* have created an invalid selector, don't update it
+			// Note that there may be false alarms if selector uses jQuery extensions
+			try {
+				document.querySelector( selector );
+				args[ 0 ] = selector;
```
gibson042: I like this "do no harm" approach, and the distinct warnings for fixable vs. unfixable selectors. But maybe we should only output one message for each bad selector, by moving the above warning to the end of this `try` block.

src/core.js
```diff
@@ -12,11 +12,31 @@ jQuery.fn.init = function( selector ) {
...
 		// The nonstandard and undocumented unquoted-hash was removed in jQuery 1.12.0
-		// Note that this doesn't actually fix the selector due to potential false positives
-		migrateWarn( "Attribute selectors with '#' must be quoted: '" + selector + "'" );
+		// First see if qS thinks it's a valid selector, if so avoid a false positive
+		try {
+			document.querySelector( selector );
+		} catch ( err1 ) {
+
+			// Didn't *look* valid to qSA, warn and try quoting what we think is the value
+			migrateWarn( "Attribute selectors with '#' must be quoted: `" + selector + "`" );
```
gibson042: I think we can omit wrapping backticks for messages like this that end with the variable data.

src/core.js
```diff
@@ -12,11 +12,31 @@ jQuery.fn.init = function( selector ) {
...
+
+			selector = selector.replace( rattrHash, function( _, attr, op, value ) {
```
gibson042: This should be inside the `try` block, since `selector` might not be a string.

src/core.js
```diff
@@ -1,7 +1,7 @@
 
 var oldInit = jQuery.fn.init,
 	oldIsNumeric = jQuery.isNumeric,
-	rattrHash = /\[\s*\w+\s*[~|^$*]?=\s*(?![\s'"])[^#\]]*#/;
+	rattrHash = /\[(\s*[-\w]+\s*)([~|^$*]?=)\s*([-\w#]+)\s*\]/g;
```
gibson042: This change opened up matches that don't include `#` characters. I think the last capture group should be `([-\w#]*?#[-\w#]*)`.

src/core.js
dmethvin: Yeah that sounds cleaner, one message saying whether we could fix it or not.

src/core.js
dmethvin: Nice catch thanks!

src/core.js
dmethvin: I'll rebase against master which has a guard against non-string inputs.

dmethvin:
Rebased against current master and fixed items mentioned by @gibson042, plus added more unit tests. In the process I found a bizarre problem in PhantomJS where it sometimes fails to find any matches on a `//g` RegExp. I've hacked something in for the moment.

Anyway I also realized I have this shim in the wrong place, it should be shimming `jQuery.find()` which is called in `jQuery.fn.find()` and `jQuery()`. It should just be a case of moving the code tho.

src/core.js
```diff
@@ -13,11 +14,31 @@ jQuery.fn.init = function( selector ) {
...
+				selector = selector.replace( rattrHashGlob, function( _, attr, op, value ) {
+					return "[" + attr + op + "\"" + value + "\"]";
+				} );
+
+				// If the regexp *may* have created an invalid selector, don't update it
+				// Note that there may be false alarms if selector uses jQuery extensions
+				try {
+					document.querySelector( selector );
+					args[ 0 ] = selector;
+					migrateWarn( "Attribute selector with '#' must be quoted: " + selector );
```
gibson042: `selector` → `args[ 0 ]`

dmethvin: Yep, thanks!

PR #186, opened 2016-04-29
Serialize: Warn and fill for jQuery.param() use of ajax traditional flag

### LINKED ISSUE #173: Serialize: Warn and fill jQuery.param( x, jQuery.ajaxSettings.traditional )
Per https://github.com/jquery/jquery/pull/3030, the docs said that we weren't looking at `jQuery.ajaxSettings.traditional` as of version 1.8 but actually we were. Note that *internal* `jQuery.ajax()` calls still pass this to `jQuery.param()` which is fine, but the goal was to avoid this method having a back-door dependency on an ajax setting.

dmethvin:
Fixes #173

### DIFF
--- a/test/serialize.js
+++ b/test/serialize.js
@@ -1,0 +1,51 @@
+
+module( "serialize" );
+
+QUnit.test( "jQuery.param( x, traditional)", function( assert ) {
+	assert.expect( 12 );
+
+	var savedTraditional = jQuery.ajaxSettings.traditional,
+		data = { a: [ 1, 2 ] },
+		standardResult = "a%5B%5D=1&a%5B%5D=2",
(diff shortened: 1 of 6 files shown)


test/serialize.js
```diff
@@ -0,0 +1,51 @@
...
+			jQuery.param( data, false ), standardResult,
+			"explicit false, traditional default" );
+	} );
+
+	jQuery.ajaxSettings.traditional = true;
+
+	expectWarning( "default, traditional true", function() {
+		assert.equal(
+			jQuery.param( data ), traditionalResult,
+			"default, traditional true	" );
```
dmethvin: Just noticed the extra spaces, I'll remove before landing.

PR #188, opened 2016-05-05
Event: Warn and fill jQuery.event.props and .fixHooks

### LINKED ISSUE #187: Warn and fill jQuery.event.props and jQuery.event.fixHooks
Per discussion in https://github.com/jquery/jquery/pull/2860

dmethvin:
Fixes #187 

I've updated the warnings file with the assumption that https://github.com/jquery/jquery/issues/3104 happens which I strongly recommend.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -1,4 +1,37 @@
-var oldLoad = jQuery.fn.load;
+var oldLoad = jQuery.fn.load,
+	originalFix = jQuery.event.fix;
+
+// Skip this when testing older 3.0 betas
+if ( jQuery.event.addProp ) {
+
+	jQuery.event.props = [];
+	jQuery.event.fixHooks = {};
+
+	jQuery.event.fix = function( originalEvent ) {
+		var event,
+			type = originalEvent.type,
+			fixHook = this.fixHooks[ type ],
+			props = jQuery.event.props;
+
+		if ( props.length ) {
+			migrateWarn( "jQuery.event.props are deprecated and removed" );
+			while ( props.length ) {
+				jQuery.event.addProp( props.pop() );
+			}
+		}
+
+		if ( fixHook && ( props = fixHook.props ) && props.length ) {
+			migrateWarn( "jQuery.event.fixHooks are deprecated and removed" );
+			while ( props.length ) {
+			   jQuery.event.addProp( props.pop() );
+			}
+		}
+
+		event = originalFix.call( this, originalEvent );
+
+		return fixHook && fixHook.filter ? fixHook.filter( event, originalEvent ) : event;
+	};
+}
 
 jQuery.each( [ "load", "unload", "error" ], function( _, name ) {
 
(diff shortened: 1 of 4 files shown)


src/event.js
```diff
@@ -1,4 +1,37 @@
-var oldLoad = jQuery.fn.load;
+var oldLoad = jQuery.fn.load,
+	originalFix = jQuery.event.fix;
+
+// Skip this when testing older 3.0 betas
```
mgol: testing?

dmethvin: maybe "testing with older jQuery 3.0 betas"?

warnings.md
```diff
@@ -172,3 +172,10 @@ See jQuery-ui [commit](https://github.com/jquery/jquery-ui/commit/c0093b599fcd58
 **Cause:** Calling `.toggleClass()` with no arguments, or with a single Boolean `true` or `false` argument, has been deprecated. Its behavior was poorly documented, but essentially the method saved away the current `class` value in a data item when the class was removed and restored the saved value when it was toggled back. If you do not believe you are specificially trying to use this form of the method, it is possible you are accidentally doing so via an inadvertent undefined value, as `.toggleClass( undefined )` toggles all classes.
 
 **Solution:** If this functionality is still needed, save the current full `.attr( "class" )` value in a data item and restore it when required.
+
+### JQMIGRATE: jQuery.event.props are deprecated and removed
+### JQMIGRATE: jQuery.event.fixHooks are deprecated and removed
```
mgol: What about `jQuery.event.mouseHooks` & `jQuery.event.keyHooks`?

src/event.js
mgol: Hmm, that would be better but maybe `checking`? Maybe that's just me but when I see `testing` in a comment I think I'm in the test code, not source. :)

dmethvin: SGTM

warnings.md
dmethvin: We never documented them, although we did [document these props and fixHooks](https://learn.jquery.com/events/event-extensions/). It would be possible to add them but I'd prefer to wait until we get some sign that people were modifying them since it will add some more overhead for every event to see if anything was added.

mgol: Hmm, OK then, I wasn't aware of a different documentation level between those APIs. I was thinking we might at least mention they're removed as well but if they weren't documented that's not necessary.

src/event.js
```diff
@@ -1,4 +1,37 @@
...
+
+		if ( props.length ) {
+			migrateWarn( "jQuery.event.props are deprecated and removed" );
+			while ( props.length ) {
+				jQuery.event.addProp( props.pop() );
+			}
+		}
+
+		if ( fixHook && ( props = fixHook.props ) && props.length ) {
+			migrateWarn( "jQuery.event.fixHooks are deprecated and removed" );
```
jbedard: This will only be done if `fixHooks[event].props` is used. What if only `fixHooks[event].filter` is used without any `props`?

dmethvin: That should be handled below in the `return`, right?

jbedard: Yeah, but right now the warning is only done if `.props` is used, not if `.filter` is used without `.props`. Is that unlikely?

dmethvin: Oh duh, now I see your point. It certainly could happen, I'll open a ticket for it. Thanks!

PR #191, opened 2016-05-13
Tests: Refactor iframe tests to be more like core

dmethvin:
This makes the arg order and basic approach closer to what is now being used in
core. It also removes some of the boilerplate from the iframe html pages and
moves it to a new iframeTest.js file that they all include.


dmethvin:
I landed this because I have a few other things that depend on it, feel free to comment if you think it needs work!

PR #192, opened 2016-05-13
Event: Warn on fixHooks case when it has no props and only filter

### LINKED ISSUE #190: Ensure touch/pointer fixhook plugins are correctly handled by Migrate
Add a unit test to be sure that more extensive fixHook plugins are actually treated properly.

dmethvin:
Fixes #190

The single hooks iframe test was getting hairy so the jQuery.event.props test was moved into a separate test.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -11,16 +11,19 @@ jQuery.event.fix = function( originalEvent ) {
 		props = jQuery.event.props;
 
 	if ( props.length ) {
-		migrateWarn( "jQuery.event.props are deprecated and removed" );
+		migrateWarn( "jQuery.event.props are deprecated and removed: " + props.join() );
 		while ( props.length ) {
 			jQuery.event.addProp( props.pop() );
 		}
 	}
 
-	if ( fixHook && ( props = fixHook.props ) && props.length ) {
-		migrateWarn( "jQuery.event.fixHooks are deprecated and removed" );
-		while ( props.length ) {
-		   jQuery.event.addProp( props.pop() );
+	if ( fixHook && !fixHook._migrated_ ) {
+		fixHook._migrated_ = true;
+		migrateWarn( "jQuery.event.fixHooks are deprecated and removed: " + type );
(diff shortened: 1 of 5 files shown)


src/event.js
```diff
@@ -11,16 +11,19 @@ jQuery.event.fix = function( originalEvent ) {
 		props = jQuery.event.props;
 
 	if ( props.length ) {
-		migrateWarn( "jQuery.event.props are deprecated and removed" );
+		migrateWarn( "jQuery.event.props are deprecated and removed: " + props.join() );
```
dmethvin: This seemed useful info, at least it's something to search for and the name gives a clue as to what might have added it.

src/event.js
```diff
@@ -11,16 +11,19 @@ jQuery.event.fix = function( originalEvent ) {
...
 			jQuery.event.addProp( props.pop() );
 		}
 	}
 
-	if ( fixHook && ( props = fixHook.props ) && props.length ) {
-		migrateWarn( "jQuery.event.fixHooks are deprecated and removed" );
-		while ( props.length ) {
-		   jQuery.event.addProp( props.pop() );
+	if ( fixHook && !fixHook._migrated_ ) {
+		fixHook._migrated_ = true;
```
dmethvin: I struggled with exactly how to do this. Alternatively I could keep some sort of flag inside Migrate as to whether I've warned yet for this fixHook.

src/event.js
```diff
@@ -11,16 +11,19 @@ jQuery.event.fix = function( originalEvent ) {
...
+		migrateWarn( "jQuery.event.fixHooks are deprecated and removed: " + type );
+		if ( ( props = fixHook.props ) && props.length ) {
```
dmethvin: Here I'm assuming the props aren't later changed, probably a safe assumption.

dmethvin:
Coverage really didn't decrease, coveralls doesn't see the tests that happen in the iframe.

Issue #15, opened 2013-01-23
Possibility to "subscribe" to warnings


dougwilson:
It would be really neat if there was a way to write a function that would be called for a warning instead of only being able to use a console or periodically check `jQuery.migrateWarnings`. Possibly a event triggered on `document` when there is a warning?

My use case is a large single page application that uses jQuery around the place. It would be nice to set `jQuery.migrateMute = true` and add a function that listened for warnings and reported them somewhere custom. Possibly something like:

``` javascript
$(document).on('warning.jqmigrate', function (event, message) {
  handleJqueryMigrateMessage(message);
});
```

PR #196, opened 2016-05-27
Add support for integrating with external reporting

sullivanst:
We use Raygun to record JavaScript errors. It would be nice to have migrate warnings there too. Raygun likes Error objects - and we like stack traces - so save Error objects in parallel with the warning messages in migrateWarn() and also raise an event that a listener can trigger on to extract the data & send it wherever it wishes.

### DIFF
--- a/src/migrate.js
+++ b/src/migrate.js
@@ -49,7 +49,7 @@ function migrateWarn( msg ) {
 	if ( !warnedAbout[ msg ] ) {
 		warnedAbout[ msg ] = true;
 		jQuery.migrateWarnings.push( msg );
-		jQuery.migrateWarnings.asErrors.push( msg );
+		jQuery.migrateWarnings.asErrors.push( new Error( msg ) );
 		if ( console && console.warn && !jQuery.migrateMute ) {
 			console.warn( "JQMIGRATE: " + msg );
 			if ( jQuery.migrateTrace && console.trace ) {
--- a/src/migrate.js
+++ b/src/migrate.js
@@ -31,6 +31,7 @@ var warnedAbout = {};
 
 // List of warnings already given; public read only
 jQuery.migrateWarnings = [];
+jQuery.migrateWarnings.asErrors = [];
 
 // Set to false to disable traces that appear with warnings
 if ( jQuery.migrateTrace === undefined ) {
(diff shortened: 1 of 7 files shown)


mgol:
What's the purpose of `jQuery.migrateWarnings.asErrors`? It just duplicates the `jQuery.migrateWarnings` array.

sullivanst:
@mgol Oops, I missed of wrapping it in a new Error() - added that now.

sullivanst:
I made the mistake of using my actual full name instead of my github username when signing the CLA. How do I fix those commits?

gibson042:
> I made the mistake of using my actual full name instead of my github username when signing the CLA. How do I fix those commits?

You can rewrite b72995efdc33240f56a9621dcd865c997bc5eeb9 with `git rebase -i b72995e^`, marking it for edit and using `git commit --amend --reset-author` (or `--author=…`) before `git rebase --continue`. Once your `sullivanst-patch-raiseEvents` history is clean, `git push --force` it to GitHub.

src/migrate.js
```diff
@@ -40,20 +41,22 @@ if ( jQuery.migrateTrace === undefined ) {
...
-	jQuery.migrateWarnings.length = 0;
+	jQuery.migrateWarnings.length = jQuery.migrateWarnings.asErrors.length = 0;
 };
 
 function migrateWarn( msg ) {
 	var console = window.console;
 	if ( !warnedAbout[ msg ] ) {
 		warnedAbout[ msg ] = true;
 		jQuery.migrateWarnings.push( msg );
+		jQuery.migrateWarnings.asErrors.push( new Error( msg ) );
```
gibson042: I agree with @mgol; this is completely redundant.

dmethvin:
I think that the discussion in #15 still covers this use case, it doesn't seem to be a common enough need to add more plumbing. The warnings are already exposed as an array, you can enumerate them as needed and send them up to a server. If your error telemetry package already traps console messages it seems like you'd be getting this for free anyway.

src/migrate.js
```diff
@@ -40,20 +41,22 @@ if ( jQuery.migrateTrace === undefined ) {
...
 		if ( console && console.warn && !jQuery.migrateMute ) {
 			console.warn( "JQMIGRATE: " + msg );
 			if ( jQuery.migrateTrace && console.trace ) {
 				console.trace();
 			}
 		}
+		jQuery( window ).trigger( "migrateWarning" );
```
dmethvin: The only time we currently call back into jQuery is to shim a method for warnings. This introduces a situation where we are using the host jQuery's functionality, which makes me nervous. It could potentially cause a recursive crash if Migrate was to give `.trigger()` warnings in the future.

gibson042: I'd consider it bad form to leak global events, but would be happy to see a hook that gets invoked here, including event emitter style (e.g., `jQuery.migrateWarningHooks.push( handler )` or `jQuery.migrateWarnings.on( "warning", handler )`).

sullivanst: I have to say I hadn't thought of intercepting the console. Obviously that would require unmuting jquery-migrate, and adding the wrapper in that intercept would be another uninformative line in the stack trace, but that's certainly no dealbreaker. Would scraping for /^JQMIGRATE:/ be safe both now and in the future for consumers who have other cruft in their console they aren't interested in recording?

I understand your hesitation about looping back through .trigger, although it would certainly be possible if you _were_ duckpunching trigger to call the original directly in that one case (or presumably if you were raising warnings, there's a decent chance it'd be because some other way would be preferred).

src/migrate.js
sullivanst: new Error( msg ).stack makes a difference (I managed to forget to put the wrapping in at first). We have a warning I'll have to try to track down after the weekend because we're muted for production and migrateWarnings doesn't provide context.

src/migrate.js
gibson042: > I have to say I hadn't thought of intercepting the console. Obviously that would require unmuting jquery-migrate, and adding the wrapper in that intercept would be another uninformative line in the stack trace

Then duckpunch `jQuery.migrateWarnings.push` like I suggested in [#15](https://github.com/jquery/jquery-migrate/issues/15#issuecomment-12632266).

sullivanst: That'll work.

Issue #98, opened 2015-03-25
Publish to npm


wbinnssmith:
Given that [npm is now recommended for jquery plugins](http://plugins.jquery.com/), it makes sense that jquery-migrate be published to npm to support the myriad of outdated plugins out there.

Ideally a `main` field is added to the package.json here and a `prepublish` step runs the usual build process. A `.npmignore` file could optionally be used to avoid publishing the sources and tests to the registry.

I'd love to submit a PR for the second half -- let me know if you're open. Of course publishing would be up to the jquery project :smile:

Edit: Just realize that `jquery-migrate` will break as-is with most commonjs bundlers and the default export in jquery (which doesn't leak `$` or `jQuery` to the window). A UMD-style wrapper similar to what jquery uses would probably be best. Likewise, I'd love to submit a pr for this too.

dmethvin:
> Just realize that jquery-migrate will break as-is with most commonjs bundlers and the default export in jquery (which doesn't leak $ or jQuery to the window).

We should separate the "Publish to npm" issue from the "Support CJS internally" one. They're not related. 

Clearly right now we need to serve the 80% (probably more like 95%) case where jQuery is a global and no module loader is being used. As long as we don't *break* that case I am fine with handling all the others and would welcome PRs!

Atsman:
Hello,

When do you plan to publish to npm? 

Best regards,
A;eh

dmethvin:
Publishing was added to the release script in e3593e22bf90008f12754b8b8fc91bd5c9d513c3 and we'll be publishing 1.3.0 within the next couple of days.

PR #206, opened 2016-07-07
Build: add UMD support

niksy:
This PR implements UMD support as discussed in #98 regarding "Support CJS internally".

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,9 @@
 		"test": "grunt test",
 		"ci": "grunt ci"
 	},
+	"dependencies": {
+		"jquery": "3.0.0"
+	},
 	"devDependencies": {
 		"chalk": "1.1.3",
 		"commitplease": "2.3.1",
--- a/src/intro.js
+++ b/src/intro.js
@@ -1,2 +1,14 @@
-(function( jQuery, window ) {
+;(function( factory ) {
+	/* jshint browserify:true */
+	if ( typeof define === "function" && define.amd ) {
+		// AMD. Register as an anonymous module.
+		define( [ "jquery" ], window, factory );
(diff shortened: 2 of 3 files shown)


dmethvin:
Awesome, thanks!

dmethvin:
We may want to add `return jQuery` in outro.js so that it will return the modified jQuery when used with CJS. It's going to actually modify its incoming `jQuery` arg, but it seems like using the return value might be more natural. What do you think?

niksy:
@dmethvin yeah, that makes sense! Do you think these change should be applied to `1x-stable` branch, and not only to jQuery v3?

dmethvin:
Yes, I can cherry pick it into that branch as well and do new releases for both.

niksy:
OK, should I squash those 2 commits before merge?

dmethvin:
I can squash them when I land it.

mgol:
Why is `jquery` a dependency and not a peerDependency? This seems wrong as Migrate should be always patching an existing jQuery copy and not provide its own.

dmethvin:
I remember thinking about this at the time, the issue with peerDependencies is that npm3 requires you to install them manually and I didn't wan't to do that. Maybe there is another way out? Perhaps just devDependency, since the Migrate code itself dynamically checks that there is a jQuery core loaded.

mgol:
> I remember thinking about this at the time, the issue with peerDependencies is that npm3 requires you to install them manually and I didn't wan't to do that.

There's no way around that. The purpose of `peerDependencies` is to depend on a package that needs to have one instance and you can't create your own. The correct way to go is to put it both in `devDependencies` and `peerDependencies` (the version range in `peerDependencies` should be more forgiving) and remove from `dependencies`.

PR #208, opened 2016-07-12
Docs: Clarify wording around deprecated/removed methods

### LINKED ISSUE #204: "jQuery.expr[":"] is now jQuery.expr.pseudos" warning: unclear whether breaking change or just deprecated
According to https://github.com/jquery/jquery-migrate/issues/203#issuecomment-230646733

> In Migrate there are messages saying that some features are deprecated (may be removed in the future) and some saying that they are already removed.

This warning:
```
jQuery.expr[":"] is now jQuery.expr.pseudos
```
doesn't explicitly used either the word "deprecated" or "removed". From the abovementioned comment I kind of get that if something "is now" something else it's probably a breaking change (though jQueryUI triggers this warning and doesn't break). 

Should be more explicit.

If there *are* two classes of warnings, you should use two very precise phrases with keywords that are consistently present in each of them so that they can be searched reliably, besides being 100% unambiguous.

dmethvin:
Fixes #204


gibson042:
LGTM.

PR #211, opened 2016-07-19
Fix: Name might be other type than string

sickboy:
So don't try to camelCase a non-string.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -4,7 +4,7 @@ jQuery.data = function( elem, name, value ) {
 	var curData;
 
 	// If the name is transformed, look for the un-transformed name in the data object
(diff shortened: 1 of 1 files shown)


mgol:
`jQuery.data` second parameter must be a string according to the docs: https://api.jquery.com/jQuery.data/

Also, we almost never use `instanceof` in the jQuery code and in those rare cases that we do we check against `jQuery` or `jQuery.Event`. One of the reasons is that `instanceof` doesn't work accross realms.

mgol:
In the future, please also remember that each pull request should have an associated issue and tests that check if the issue is fixed. We're very reluctant to accept fixes without tests.

PR #212, opened 2016-07-22
Core: Use migrateWarnFunc for warning on function call, not access

dmethvin:
Unit tests already were using calls, so no changes were needed there.

migrateWarnFunc will be used in a lot of the 3.2 deprecations.


dmethvin:
Commit message should say `migrateWarnFunc` not `migrateWarnProp`.

mgol:
LGTM

PR #216, opened 2016-08-03
Effects: Warn and fill jQuery.fx.interval

### LINKED ISSUE #215: Warn and fill jQuery.fx.interval
https://github.com/jquery/jquery/issues/3154

Although we don't technically need to fill it yet it's trivial so we might as well do it.

dmethvin:
Fixes #215

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -16,3 +16,12 @@ jQuery.Tween.prototype.run = function( percent ) {
 
 	oldTweenRun.apply( this, arguments );
 };
+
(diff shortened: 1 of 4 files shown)


mgol:
LGTM.

mgol:
Or maybe we could add a test ensuring no warnings happen for browsers not having rAF?

dmethvin:
I refactored the tests a bit to assert that non-rAF browsers don't warn. Coverage decreases since the tests are conditioned on the presence of rAF. I checked manually on IE9 and Android and it reports correctly.

PR #222, opened 2016-08-29
Traces: commas replaced by semi-colons

LaurentBarbareau:
Standardization of the trace text shape to facilitate textual search into https://github.com/jquery/jquery-migrate/blob/1.x-stable/warnings.md
I arbitrary chose the semi-colon as a sentence separator. This is not important, the goal is to be homogeneous between the sources and the documentation, because there is at least one inconsistency with "JQMIGRATE: jQuery.parseJSON is deprecated; use JSON.parse", that I want to fix too.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -62,7 +62,7 @@ for ( findProp in oldFind ) {
 
 // The number of elements contained in the matched element set
 jQuery.fn.size = function() {
-	migrateWarn( "jQuery.fn.size() is deprecated and removed, use the .length property" );
+	migrateWarn( "jQuery.fn.size() is deprecated and removed; use the .length property" );
 	return this.length;
 };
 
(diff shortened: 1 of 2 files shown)


mgol:
It's semi-colons, not semi-columns. :-) I agree we should be consistent; which style do you prefer, @dmethvin?

dmethvin:
This change with semicolons is fine with me. @LaurentBarbareau can you also check that this is consistent with the `warnings.md` file?

LaurentBarbareau:
Ok, thank you. I've done the modifications into warnings.md https://github.com/jquery/jquery-migrate/pull/223 but some commas still need to be replaced by semi-colons into that file and the corresponding sources ;)

dmethvin:
LGTM! Thanks @LaurentBarbareau we'll land this soon.

PR #223, opened 2016-08-29
Warning message: commas replaced by semi-colons

LaurentBarbareau:
Consistency with https://github.com/jquery/jquery-migrate/pull/222/commits/cfc0f4ef1028a4d9e423e2ace258a5355b875c3d

### DIFF
--- a/warnings.md
+++ b/warnings.md
@@ -79,7 +79,7 @@ This is _not_ a warning, but a console log message the plugin shows when it firs
 
 **Solution**: Replace any use of `.andSelf()` with `.addBack()`.
 
-### JQMIGRATE: jQuery.fn.size() is deprecated and removed, use the .length property
+### JQMIGRATE: jQuery.fn.size() is deprecated and removed; use the .length property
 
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR but let's keep all the related changes in #222; we'd like to land them in one commit anwyay.

dmethvin:
@LaurentBarbareau Just make another commit on your `patch-1` branch over in #222 and push it. When we land it we will squash the two commits. The reason for keeping it in one place is to be sure that the master branch is consistent as often as possible rather than having the work be half-done in two different commits. Thanks for your help, if you need assistance with this just post a question in #222!

LaurentBarbareau:
I understand why you want to keep all in #222 but I don't know how to do.
Can I make it from the GitHub web interface?
thx

dmethvin:
I didn't know the answer before but just found out. :smile_cat: 

Go to the jquery fork in your account and find the `patch-1` branch, it's at https://github.com/LaurentBarbareau/jquery-migrate/tree/patch-1 . Then go to the file you want to edit and click the little pencil icon  Make the change and at the bottom it should have "Commit directly to the patch-1 branch". That will add another commit to your pull request!

dmethvin:
Normally the Github web interface isn't flexible enough, since it isn't possible to create and run additional unit tests that way. For documentation issues like this it is fine though.

LaurentBarbareau:
Thank you.
Done.

PR #226, opened 2016-08-31
Build: test with jQuery 3.0.0 & 3.1.0

mgol:


dmethvin:
LGTM. We need some alert for this each time a new core version arrives!

mgol:
I reordered the entries (the newest should come on top) and landed the PR.

PR #228, opened 2016-09-26
Docs: Remove reference to deprecated `.ready()` method

### LINKED ISSUE #221: In warnings.md shows 'ready' event is deprecated but lists as $(document).ready(func); as solution that also is depricated
The document [Warnings.md](https://github.com/jquery/jquery-migrate/blob/master/warnings.md) have information about ready being depricated that the plugin also warns about.

But it also notes following as a solution.

``` javascript
$( document ).ready( fn )
```

Looking at [Upgrade-Guide](http://jquery.com/upgrade-guide/3.0/#deprecated-document-ready-handlers-other-than-jquery-function) It notes that the solution above is removed. I can also confirm that the migrate plugin does not warn about this.

dmethvin:
Fixes #221


dmethvin:
Landed at 4d60965afceb8379bd71c7dcfec078e24678dde7

PR #229, opened 2016-09-27
Build: Followup to #206, fix jQuery dependency

dmethvin:
Is this along the lines of what you were thinking in https://github.com/jquery/jquery-migrate/pull/206#issuecomment-246420065 @mgol? Since we are pulling the actual dependencies from the CDN for testing, we don't actually use these npm-installed copies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
 		"test": "grunt test",
 		"ci": "grunt ci"
 	},
-	"dependencies": {
-		"jquery": "3.0.0"
+	"peerDependencies": {
+		"jquery": ">3.0.0"
 	},
 	"devDependencies": {
 		"chalk": "1.1.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -22,8 +22,8 @@
 		"test": "grunt test",
 		"ci": "grunt ci"
 	},
-	"dependencies": {
-		"jquery": "3.0.0"
+	"peerDependencies": {
+		"jquery": ">3.0.0"
```
dmethvin: True, and I wonder if I should also limit to `<4.0.0` since we'll plan on a new round of removals then?

mgol:
Yep, I was thinking about specifying ">=3 <4". That should be enough.

-- 
Michał Gołębiowski

PR #232, opened 2016-10-04
Fill in missing link

maxnordlund:
Just noticed this when reading the README, and thought why not just fix it. 😸


dmethvin:
Nice! thanks. 😸

Issue #233, opened 2016-10-24
Easing compatibility sending initial percent value (0) to custom easing function


interacsean:
When older, third-party easing definitions are used (i.e. which use the deprecated, multiparameter easing format), jquery-migrate would send the initial animation percent value (of 0) to the easing function on each render tick, rather than the actual progressive percent value.  This causes any animations using these third-party easing definitions to be non-responsive.

Demo here: http://codepen.io/interacsean/pen/KgQgEb

Failing test and fix coming through shortly...

PR #234, opened 2016-10-24
Bugfix/easing compatibility

interacsean:
Addresses bug #233 by listening for the current percent that the easing function should be called with, rather than using the percent passed to jQuery.Tween.prototype.run, which will always be 0.

The expanded test asserts that calls to the easing function come through with 0 the first time, >0 the second time, and after the animation has completed, that it was most recently called with 1.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -1,6 +1,6 @@
 var oldTweenRun = jQuery.Tween.prototype.run;
 
-jQuery.Tween.prototype.run = function( percent ) {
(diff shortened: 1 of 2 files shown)


dmethvin:
Hey @interacsean can you sign our new CLA? Sorry about the delay in landing this.

interacsean:
@dmethvin I've signed that now, though I noticed my github email address is different to the one which is against my commits.  Hope that doesn't confuse the bots...

Issue #198, opened 2016-06-21
Naive shim around $.data breaks much code


markrendle:
The shim around .data doesn't support the same set of overloads as the core function. It assumes the second parameter is the `name` property, whereas it can be an object to set a bunch of properties simultaneously.

See: https://github.com/jquery/jquery-migrate/blob/master/src/data.js#L7
vs: https://github.com/jquery/jquery/blob/master/src/data.js

Among other things, this breaks AngularJS.

dmethvin:
There is a difference between `jQuery.data()`, which Migrate shims, and `jQuery.fn.data()` which Migrate does not touch. Can you provide a test case that shows the problem?

dmethvin:
Closing, but if you have some more information I can reopen.

mgol:
This seems to be breaking Angular: https://github.com/angular/angular.js/issues/15370. Let's reconsider.

mgol:
There is a PR for this issue at #235.

mgol:
@markrendle Could you make sure [the development version of Migrate](https://code.jquery.com/jquery-migrate-git.js) fixes your issue?

donpisci:
@mgol Have you any idea when you think the version with the 'string.replace is not a function' fix will be officially released?

Thanks!

dmethvin:
@donpisci The PR was #235 and landed for the 3.0.1 release. Are you still seeing the problem with that version? If so please open a new issue and provide a test case.

donpisci:
@dmethvin Yes, I found the 3.0.1 release shortly after leaving the message and hadn't got round to updating my message- apologies! 

All's working now.

PR #235, opened 2016-10-27
Data: Support object as second argument

wjmao88:
jQuery.data supports passing an object as a second argument and sets data for each entry in that object.

This warns the user about each violating non-camelCase key in the object.

This also adds a type check to avoid a current bug where jQuery.camelCase throws an error because a non-string is passed to it.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2,9 +2,21 @@ var oldData = jQuery.data;
 
 jQuery.data = function( elem, name, value ) {
 	var curData;
+	
+	//name can be an object, and each entry in the object is meant to be set as data
+	if ( name && typeof name === 'object' && arguments.length === 2 ) {
+		for ( var key in name ) {
+			if ( key !== jQuery.camelCase( key ) ){
+				migrateWarn( "jQuery.data() always sets/gets camelCased names: " + key );
+			}
+		}
+		
+		//jQuery.data will convert keys to camelCase, and this is a setter.	
+		return oldData.apply( this, arguments );
+	}
 
 	// If the name is transformed, look for the un-transformed name in the data object
-	if ( name && name !== jQuery.camelCase( name ) ) {
+	if ( name && typeof name === 'string' && name !== jQuery.camelCase( name ) ) {
 		curData = jQuery.hasData( elem ) && oldData.call( this, elem );
 		if ( curData && name in curData ) {
 			migrateWarn( "jQuery.data() always sets/gets camelCased names: " + name );
--- a/src/data.js
+++ b/src/data.js
@@ -2,16 +2,16 @@ var oldData = jQuery.data;
 
 jQuery.data = function( elem, name, value ) {
 	var curData;
-	
-	//name can be an object, and each entry in the object is meant to be set as data
+
+	//Name can be an object, and each entry in the object is meant to be set as data
 	if ( name && typeof name === "object" && arguments.length === 2 ) {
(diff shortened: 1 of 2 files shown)


src/data.js
```diff
@@ -3,8 +3,26 @@ var oldData = jQuery.data;
 jQuery.data = function( elem, name, value ) {
 	var curData;
 
+	//Name can be an object, and each entry in the object is meant to be set as data
```
mgol: There should be a space after `//`.

src/data.js
```diff
@@ -3,8 +3,26 @@ var oldData = jQuery.data;
...
+		for ( var key in name ) {
+			if ( key !== jQuery.camelCase( key ) ) {
+				migrateWarn( "jQuery.data() always sets/gets camelCased names: " + key );
+				curData[ key ] = name[ key ];
+			} else {
+				sameKeys[ key ] = name[ key ];
+			}
+		}
+
+		oldData.call( this, elem, sameKeys );
```
mgol: Why not `oldData.apply( this, arguments )` similarly to how what the other configuration is handled? You wouldn't have to record `sameKeys` at all then.

test/data.js
```diff
@@ -37,6 +37,38 @@ test( "jQuery.data() camelCased names", function( assert ) {
 		} );
 	} );
 
+	div = document.createElement( "div" );
```
mgol: Could you move all this code to a separate `test`? An example name: "jQuery.data() camelCased names (mass setter)".

mgol:
@wjmao88 Sorry for keeping you wait this long with the review. Will you have time to address my comments? If not, I can finish it for you.

mgol:
This PR should fix issue #198.

src/data.js
wjmao88: sameKeys is not the same as the second argument, as it only contains keys that are in camelCase and thus does not need to be handled here. 

The other oldData.apply is actually doing the same if it determines the key is already camel case.

wjmao88:
@mgol Yes, I'll be able to find time within a day or so.

src/data.js
```diff
@@ -3,8 +3,26 @@ var oldData = jQuery.data;
...
+
+		return name;
+	}
+
 	// If the name is transformed, look for the un-transformed name in the data object
-	if ( name && name !== jQuery.camelCase( name ) ) {
+	if ( name && typeof name === "string" && name !== jQuery.camelCase( name ) ) {
```
mgol: I don't think this change is necessary. We only support documented scenarios, while we want to add support for the object as the second parameter here at this point we know it's not an object so we're fine.

src/data.js
mgol: Ah, right, I was comparing it to the second part but there's an early return there if the non-camelCased version of the key was found. It's fine then. 👍

src/data.js
wjmao88: I agree with your reason, which is why I think it should goes straight to jQuery's own .data method, and will throw the same error as when jquery-migrate is not used.

dmethvin: @mgol I see @wjmao88 's point here, it's ensuring the behavior (and error location) is identical in both cases. Are you good with that?

mgol:
Yes, his reasoning sounds reasonable!

mgol:
@wjmao88 Landed, thanks!

PR #238, opened 2016-11-02
Build: Switch from JSHint/JSCS to ESLint (+ other housekeeping)

### LINKED ISSUE #236: Switch to ESLint
We should switch from JSHint+JSCS to ESLint as jQuery Core has.

mgol:
Fixes #236

It makes sense to review it commit by commit, the changes make sense to land separately.


mgol:
ping @dmethvin. :) @timmywil gave an LGTM.

package.json
```diff
@@ -1,80 +1,81 @@
 {
-	"name": "jquery-migrate",
```
mgol: Why? It already says "2 spaces" so this file was incorrectly formatted...

markelog: O, can't really see tabs vs spaces thingy in github UI

Gruntfile.js
```diff
@@ -136,23 +118,15 @@ module.exports = function( grunt ) {
...
+	require( "load-grunt-tasks" )( grunt );
 
 	// Integrate jQuery migrate specific tasks
 	grunt.loadTasks( "build/tasks" );
 
 	// Just an alias
 	grunt.registerTask( "test", [ "qunit" ] );
 
-	grunt.registerTask( "lint", [ "jshint", "jscs" ] );
+	grunt.registerTask( "lint", [ "eslint:dev", "eslint:dist" ] );
```
markelog: Why two tasks? Why not just `eslint`?

src/migrate.js
```diff
@@ -1,11 +1,14 @@
+/* exported migrateWarn, migrateWarnFunc, migrateWarnProp */
```
markelog: What this means?

Gruntfile.js
mgol: This is one task with 2 targets. Not specifying the targets manually would first invoke the `dist` one and then quit; that makes it way harder to know what to fix.

markelog: > Not specifying the targets manually would first invoke the dist one and then quit

How this is an issue?

> that makes it way harder to know what to fix

Didn't it prints path to problematic file?

mgol:
The problematic file in question would always be dist/jquery-migrate.js.
Hardly a good experience! I want this file to be pointed at only if we
already verified that source files pass the linter.

markelog:
@mgol Make sense

markelog:
Don't be so defensive btw :)

mgol:
I wasn't trying to be defensive, sorry if it sounded that way. :)

PR #241, opened 2016-11-28
Core: Don't use .apply() on console.log to satisfy IE9

### LINKED ISSUE #231: [IE9] window.console.log.apply - Object doesn't support property or method 'apply'
After load page, i see this error.

problem in `src/migrate.js`;
```js
	// Support: IE9 only
	// IE9 only creates console object when dev tools are first opened
	// Also, avoid Function#bind here to simplify PhantomJS usage
	var log = window.console && window.console.log &&
			function() { window.console.log.apply( window.console, arguments ); }, // <------ this
		rbadVersions = /^[12]\./;
```

related [questions](http://stackoverflow.com/questions/5538972/console-log-apply-not-working-in-ie9) on stackoverflow

dmethvin:
Fixes #231 

Not sure of a good way to test this for IE9, which is why the problem existed in the first place.

### DIFF
--- a/src/migrate.js
+++ b/src/migrate.js
@@ -2,29 +2,25 @@
 
 ( function() {
 
+	var rbadVersions = /^[12]\./;
+
 	// Support: IE9 only
 	// IE9 only creates console object when dev tools are first opened
-	// Also, avoid Function#bind here to simplify PhantomJS usage
-	var log = window.console && window.console.log &&
-		function() {
-			window.console.log.apply( window.console, arguments );
-		},
-		rbadVersions = /^[12]\./;
-
-	if ( !log ) {
+	// IE9 console is a host object, callable but doesn't have .apply()
+	if ( !window.console || !window.console.log ) {
 		return;
 	}
 
 	// Need jQuery 3.0.0+ and no older Migrate loaded
 	if ( !jQuery || rbadVersions.test( jQuery.fn.jquery ) ) {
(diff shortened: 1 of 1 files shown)


mgol:
LGTM. I don't see a way to test it in IE 9 but we could simulate its situation via:
```js
if ( !window.console ) {
	window.console = {};
}
if ( !window.console.log ) {
	window.console.log = function () {};
	window.console.log.__proto__ = null;
}
```

src/migrate.js
```diff
@@ -2,29 +2,25 @@
...
-	// Also, avoid Function#bind here to simplify PhantomJS usage
-	var log = window.console && window.console.log &&
-		function() {
-			window.console.log.apply( window.console, arguments );
-		},
-		rbadVersions = /^[12]\./;
-
-	if ( !log ) {
+	// IE9 console is a host object, callable but doesn't have .apply()
+	if ( !window.console || !window.console.log ) {
```
markelog: `typeof window.console.apply` wouldn't work?

mgol: It won't as there is no `window.console` in IE 9.

dmethvin: Right, and even when `window.console` exists in IE9 (when the debugger window is active) it is a "host object" that is callable but doesn't have `.apply()`.

dmethvin:
@markelog @mgol same for this PR

PR #243, opened 2016-12-05
Event: Warn and fill event aliases

### LINKED ISSUE #230: Warn and fill event shorthands
Core ref https://github.com/jquery/jquery/issues/3214
Docs ref https://github.com/jquery/api.jquery.com/issues/972

dmethvin:
Fixes #230

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -75,6 +75,20 @@ jQuery.each( [ "load", "unload", "error" ], function( _, name ) {
 
 } );
 
+jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
+	"mousedown mouseup mousemove mouseover mouseout mouseenter mouseleave " +
+	"change select submit keydown keypress keyup contextmenu" ).split( " " ),
+	function( i, name ) {
+
+	// Handle event binding
+	jQuery.fn[ name ] = function( data, fn ) {
+		migrateWarn( "jQuery.fn." + name + "() event shorthand is deprecated" );
+		return arguments.length > 0 ?
+			this.on( name, null, data, fn ) :
+			this.trigger( name );
+	};
+} );
+
 // Trigger "ready" event only once, on document ready
 jQuery( function() {
 	jQuery( window.document ).triggerHandler( "ready" );
@@ -107,5 +121,9 @@ jQuery.fn.extend( {
 		return arguments.length === 1 ?
 			this.off( selector, "**" ) :
 			this.off( types, selector || "**", fn );
+	},
+	hover: function( fnOver, fnOut ) {
+		migrateWarn( "jQuery.fn.hover() is deprecated" );
+		return this.on( "mouseenter", fnOver ).on( "mouseleave", fnOut || fnOver );
 	}
 } );
--- a/test/event.js
+++ b/test/event.js
@@ -95,25 +95,27 @@ QUnit.test( ".delegate() and .undelegate()", function( assert ) {
 } );
 
 QUnit.test( "Event aliases", function( assert) {
-	assert.expect( 16 );
+	assert.expect( 14 );
 
 	var $div = jQuery( "<div />" );
 
 	"scroll click submit keydown".split( " " ).forEach( function( name ) {
 		expectWarning( "." + name + "()", 1, function() {
 			$div[ name ]( function( event ) {
-				assert.equal( event.type, name, name)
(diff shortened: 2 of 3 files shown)


warnings.md
```diff
@@ -199,3 +199,15 @@ See jQuery-ui [commit](https://github.com/jquery/jquery-ui/commit/c0093b599fcd58
 **Cause:** The calling code has attempted to attach a `load` event to `window` after the page has already loaded. That means the handler will never run and so is probably not what the caller intended. This can occur when the event attachment is made too late, for example, in a jQuery ready handler. It can also occur when a file is loaded dynamically with jQuery after the page has loaded, for example using the `$.getScript()` method.
 
 **Solution:** If a function `fn` does not actually depend on all page assets being fully loaded, switch to a ready handler `$( fn )` which runs earlier and will aways run `fn` even if the script that contains the code loads long after the page has fully loaded. If `fn` actually does depend on the script being fully loaded, check `document.readyState`. If the value is `"complete"` run the function immediately, otherwise use `$(window).on( "load", fn )`.
+
+### JQMIGRATE: jQuery.fn.click() event shorthand is deprecated
```
markelog: Why not generalize this header?

src/event.js
```diff
@@ -75,6 +75,20 @@ jQuery.each( [ "load", "unload", "error" ], function( _, name ) {
...
 
+jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
+	"mousedown mouseup mousemove mouseover mouseout mouseenter mouseleave " +
+	"change select submit keydown keypress keyup contextmenu" ).split( " " ),
+	function( i, name ) {
+
+	// Handle event binding
+	jQuery.fn[ name ] = function( data, fn ) {
+		migrateWarn( "jQuery.fn." + name + "() event shorthand is deprecated" );
+		return arguments.length > 0 ?
```
markelog: Can we remember the old functions and call them instead duplicating logic in core?

mgol: What do you mean? This code should work even if event shorthands are removed from Core.

markelog: As far as i know we don't plan to remove it. And if/when we remove it, warning should be changed regardless

markelog: It seems reimplementing is more preferable, judging by the rest of the source anyway

warnings.md
dmethvin: I tried to figure out the best way to do this. My concern with generalizing the header was that it wouldn't be an exact match. The most common shorthand (I think) is click so i figured this approach would yield the best results with the others in the text. Another way is to duplicate the exact messages but that would get out of hand with so many on the list now. Or I can change it to something like `METHOD`. What do you think?

src/event.js
dmethvin: I think our approach for 3.0 has been to implement the deprecated functionality even if it's still in core. It does handle the scenario of custom builds better i guess.

dmethvin:
@markelog @mgol LMK if there's anything more you'd like done here.

test/event.js
```diff
@@ -94,6 +94,37 @@ QUnit.test( ".delegate() and .undelegate()", function( assert ) {
...
 
+QUnit.test( "Event aliases", function( assert) {
+	assert.expect( 16 );
+
+	var $div = jQuery( "<div />" );
+
+	"scroll click submit keydown".split( " " ).forEach( function( name ) {
+		expectWarning( "." + name + "()", 1, function() {
+			$div[ name ]( function( event ) {
+				assert.equal( event.type, name, name)
```
mgol: A missing space before `)`

mgol: ...and a missing semi-colon at the end. :)

test/event.js
```diff
@@ -94,6 +94,37 @@ QUnit.test( ".delegate() and .undelegate()", function( assert ) {
 	} );
 } );
 
+QUnit.test( "Event aliases", function( assert) {
+	assert.expect( 16 );
```
mgol: Where does the 16 come from? I counted 14:

* in the first block: 4 warnings & 4 `assert.equal`s
* in the second block: 1 warning & 2 `assert.ok`s
* in the third block: 1 warning & 2 `assert.equal`s

What am I missing?

test/event.js
dmethvin: Oh great now I have to go figure out why the eslint settings missed that.  :)

PR #247, opened 2017-01-18
Core: Warn on use of jQuery.holdReady()

### LINKED ISSUE #225: jQuery migrate jQuery.holdReady
Issue is mentioned in https://github.com/jquery/jquery/issues/3288 and pull request is https://github.com/jquery/jquery/pull/3306

dmethvin:
Fixes #225 

jQuery core includes two implementations of `jQuery.holdReady()` and it would be tough at this point to anticipate how we might rewrite it, so I skipped trying to fill it for now at least.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -275,6 +275,15 @@ test( "jQuery.expr.pseudos aliases", function( assert ) {
 
 } );
 
+QUnit.test( "jQuery.holdReady (warn only)", function( assert ) {
+	assert.expect( 1 );
+
+	expectWarning( "jQuery.holdReady", 1, function() {
+		jQuery.holdReady( true );
+		jQuery.holdReady( false );
+	} );
+} );
+
 TestManager.runIframeTest( "old pre-3.0 jQuery", "core-jquery2.html",
 	function( assert, jQuery, window, document, log ) {
 		assert.expect( 1 );
--- a/src/core.js
+++ b/src/core.js
@@ -89,6 +89,9 @@ jQuery.isNumeric = function( val ) {
 	return oldValue;
 };
 
+migrateWarnFunc( jQuery, "holdReady", jQuery.holdReady,
+	"jQuery.holdReady is deprecated" );
+
 migrateWarnFunc( jQuery, "unique", jQuery.uniqueSort,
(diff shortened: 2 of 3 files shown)


test/core.js
```diff
@@ -275,6 +275,15 @@ test( "jQuery.expr.pseudos aliases", function( assert ) {
 
 } );
 
+QUnit.test( "jQuery.holdReady (warn only)", function( assert ) {
+	assert.expect( 1 );
+
+	expectWarning( "jQuery.holdReady", 1, function() {
+		jQuery.holdReady( true );
+		jQuery.holdReady( false );
```
markelog: Why you calling it twice?

markelog:
@Krinkle I remember you were using it somewhere?

test/core.js
dmethvin: Just to be sure I don't get two warnings.

markelog: :)

Krinkle:
@markelog I used to use it at Wikipedia by calling `jQuery.ready()` at the bottom of the HTML before `</body>`. Added in 2013 with https://github.com/wikimedia/mediawiki/commit/0a000d5cf4f915f8b5cdb5e2b25674200952f4ce; Removed in 2015 with https://github.com/wikimedia/mediawiki/commit/d7905627fdc3b210a4f42d2ba23a337d0ca2bc5c.

It was mainly there to work around a bug in Firefox (as well as minor optimisation to trigger the event when its "early" enough for us, which was a bad idea in hindsight). The bug was actually caused by some fragile logic on our end, fixed and refactored in 2014, and the `jQuery.ready()` "optimisation" was removed from our code base in 2015.

For third-party users of MediaWiki, this was released in 1.22.0 (Dec 2013) and removed in 1.26.0 (Nov 2015). The current stable (1.28) and LTS (1.27) no longer have this, however we do have a few months left [until May 2017](https://www.mediawiki.org/wiki/Version_lifecycle) before MediaWiki 1.23 (legacy LTS) reaches end-of-life.

markelog:
@Krinkle thank you for the thoughtful comment, looks like we are in the clear :)

PR #249, opened 2017-02-02
Reference the final jQuery 1.x version

collinanderson:


mgol:
Landed in 84ed7b538cfc33f897411f896a2fe57bf2b336f6. Note that we don't plan further 1.x releases so this change will only be reflected on GitHub.

PR #264, opened 2017-07-20
Effects: Use linear easing when old-style easing is detected

### LINKED ISSUE #258: Animation does not run if multiple arguments is used in easing function
If easing function is added in an old way with multiple parameters, for example:

```js
$.extend($.easing, {
    easeInQuad: function (x, t, b, c, d) {
        return c*(t/=d)*t + b;
    }
 });
```

AND jQuery Migrate is included - the animation does not run at all. 

Without jQuery Migrate the animation works.

Fiddle that reproduces the issue: https://jsfiddle.net/8ng0tsyn/ (jQuery 3.2.1 and jQuery migrate 3.0.0).

A lot of libraries, unfortunately, still use "old way", the same with most popular [answer on Stackoverflow](http://stackoverflow.com/questions/5207301/jquery-easing-functions-without-using-a-plugin). And since some CMS update jQuery core and do not update jQuery plugins - this change is quite breaking on some sites.

Imo, the jQuery Migrate should fallback to default "linear" scaling if developer starts an animation with easing function that has multiple arguments. Thankfully there is [Function.length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) property.

dmethvin:
Fixes #258

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -1,17 +1,15 @@
 var oldTweenRun = jQuery.Tween.prototype.run;
+var linearEasing = function( pct ) {
+		return pct;
+	};
 
 jQuery.Tween.prototype.run = function( ) {
 	if ( jQuery.easing[ this.easing ].length > 1 ) {
 		migrateWarn(
-			"easing function " +
-			"\"jQuery.easing." + this.easing.toString() +
-			"\" should use only first argument"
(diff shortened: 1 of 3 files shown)


dmethvin:
Throwing errors on jQueries older than 3.2.1, looking into it.

Issue #254, opened 2017-03-20
Minor release for improved warnings and fixes


Krinkle:
Would be nice if a new minor release could be cut so that the improved warnings from dc1d703f7747d71f (#204) goes out.

----

The minor release from current master would contain:

* Build: add UMD support. #206
* Core: Warn on use of `jQuery.holdReady()`. #247
* **Core: Don't use `.apply()` on `console.log` to satisfy IE9.** #241
* **Data: Fix support object as second argument.** #235
* **Docs: Clarify wording around deprecated/removed methods.** #208
* **Effects: Fix for old style easing definitions compatibility.** #234
* Effects: Warn on use of `jQuery.fx.interval`. #216
* **Event: Warn and fill `jQuery.event.props.concat` for jQuery Mobile pre-1.5**. #209
* Event: Warn about late use of `$(window).on("load"...)`. #217
* Event: Use window.document to get to document. 9e3dfcbf
* Event: Warn for deprecated event aliases. #243

dmethvin:
Yeah, a 3.2 release seems like a good idea. See #220 for why isArray isn't being warned. It's a major pain to do those warnings when there are internal uses inside jQuery itself.  The positional selectors one would be really annoying since they're frequently used and I'm not sure we're ever going to remove them.

mgol:
I think we should warn only in jQuery 3.2.0+ then, better than nothing!

Krinkle:
From <https://github.com/jquery/jquery-migrate/issues/246#issuecomment-287883124>

Krinkle:
@dmethvin Added a release note draft to the start comment. Is anything still blocking a minor release?

dmethvin:
Sorry @Krinkle, I'm not sure exactly what needs to be done at the moment. I need a day to go through and see where things stand. If we're going to catch all the jQuery 3.2.0 deprecations I think there are a few things left and some may not have issues yet, but that would be part of the review.

Krinkle:
@dmethvin If we can get all jQuery 3.2.0 additions finalised soon, that'd be great. But if not, I'd like to consider not blocking on that, and perhaps have a release with a 3.0.x or 3.1.x version number first, to get these fixes released soon?

dmethvin:
SGTM, we can do another release later.

PR #268, opened 2017-09-04
Prepare 3.0.1 release

Krinkle:
Closes https://github.com/jquery/jquery-migrate/issues/254.

/cc @dmethvin


dmethvin:
@Krinkle sorry to leave things hanging so long! I'm going to land the open PRs, after that is there anything you think needs to be in 3.0.1? 

The [release script](https://github.com/jquery/jquery-migrate/blob/master/build/release.js) takes care of updating the `package.json` and `README.md` versions so we shouldn't need this PR. That script also updates the NPM packages, but the jQuery CDN has to be updated manually.

Krinkle:
Ah, okay. Thanks!

PR #290, opened 2018-02-18
Core: Warn and fill jQuery.isWindow

### LINKED ISSUE #263: Warn and fill jQuery.isWindow
https://github.com/jquery/jquery/pull/3702

dmethvin:
Fixes #263

### DIFF
--- a/warnings.md
+++ b/warnings.md
@@ -206,6 +206,11 @@ See jQuery-ui [commit](https://github.com/jquery/jquery-ui/commit/c0093b599fcd58
 
 **Solution:** Rewrite the page so that it does not require all jQuery ready handlers to be delayed. This might be accomplished, for example, by late-loading only the code that requires the delay when it is safe to run. Due to the complexity of this method, jQuery Migrate does not attempt to fill the functionality. If the underlying version of jQuery used with jQuery Migrate no longer contains `jQuery.holdReady()` the code will fail shortly after this warning appears.
 
+### JQMIGRATE: jQuery.isWindow() is deprecated
+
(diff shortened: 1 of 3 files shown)


warnings.md
```diff
@@ -206,6 +206,11 @@ See jQuery-ui [commit](https://github.com/jquery/jquery-ui/commit/c0093b599fcd58
 
 **Solution:** Rewrite the page so that it does not require all jQuery ready handlers to be delayed. This might be accomplished, for example, by late-loading only the code that requires the delay when it is safe to run. Due to the complexity of this method, jQuery Migrate does not attempt to fill the functionality. If the underlying version of jQuery used with jQuery Migrate no longer contains `jQuery.holdReady()` the code will fail shortly after this warning appears.
 
+### JQMIGRATE: jQuery.isWindow() is deprecated
+
+**Cause:** This method returns `true` if its argument is thought to be a `window` element. It was created for internal use but is no longer used by jQuery.
```
mgol: We still use the helper function internally, just not `jQuery.isWindow` directly so while technically this is true it looks like we no longer needed the utility and we removed it when in fact we've just hidden it from the public API. This was done in a larger effort of removing utility functions that we think are better either written by hand or taken from another library as we don't position jQuery as a bag of library functions.

I'd remove the last sentence from this paragraph; we could replace it with something else but I don't have very good ideas.

dmethvin:
This needs to have coded added that checks for 3.3.0 or higher, since previous versions used `.isWindow()` internally.

dmethvin:
Made requested changes and added the version check.

PR #291, opened 2018-02-18
CSS: Warn and fill jQuery.cssProps

### LINKED ISSUE #283: Warn and fill `jQuery.cssProps`
See https://github.com/jquery/jquery/issues/3653 .

dmethvin:
Fixes #283

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -11,6 +11,8 @@
 	},
 	"globals": {
 		"window": true,
+		"Proxy": true,
+		"Reflect": true,
 
 		"define": true,
 		"module": true,
--- a/src/css.js
+++ b/src/css.js
@@ -42,3 +42,13 @@ jQuery.swap = function( elem, options, callback, args ) {
 
 	return ret;
 };
+
+if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
+
+	jQuery.cssProps = new Proxy( jQuery.cssProps || {}, {
+		set: function() {
+			migrateWarn( "JQMIGRATE: jQuery.cssProps is deprecated" );
+			return Reflect.set.apply( this, arguments );
+		}
+	} );
+}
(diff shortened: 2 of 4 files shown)


dmethvin:
I'm opening a ticket for this in jQuery, it can't be landed as-is.

dmethvin:
I'm not landing this pull request. was planning on using a proxy. I have a
proof-of-concept on JSbin but I can't get to it right now because I'm on my
phone.

>
>

dmethvin:
Here's the PoC: http://jsbin.com/wewofay/edit?js,console

I still need to do a perf comparision. Hoping that getting will stay pretty fast and it doesn't really matter if setting gets slower since it shouldn't happen often at all.

dmethvin:
@Krinkle Yeah, I've updated the PR to use `Proxy` but still can't land it quite yet because it needs (at minimum) a version check which will be provided by #299. At that point it could be landed although it may make more sense to wait until core does its next release which will include the changes I have in https://github.com/jquery/jquery/pull/4005 .

dmethvin:
This shows that Proxy makes things "slower", even in the read case, but the numbers for both are so incredibly high that I'm not convinced they're right. If they are, the Proxy is certainly fast enough IMO. The most common case by far would be just reading the object so I'm not concerned about the read/write case at all, I just included it for completeness.

https://esbench.com/bench/5ab914eef2949800a0f61973

jaydott:
jquery

jaydott:
jq data

jaydott:
gdf

jaydott:
fghfg

mgol:
@dmethvin Why did you close this one?

dmethvin:
Whoops, I accidentally deleted the branch on my fork. I've got it locally and can push it back up though.

dmethvin:
Merged as 18618af56c563b5ac3da2ac56187a5acb0cd886f

PR #298, opened 2018-03-22
Effects: Don't warn about interval if document is hidden

### LINKED ISSUE #292: Incorrect warning for use of jQuery.fx.interval
Following on from #276 - I had the same problem and managed to catch it in the act. In particular, jQuery 3.2.1 does still reference `jQuery.fx.interval` in one place:

https://github.com/jquery/jquery/blob/3d732cca6b5076a9d13eee98e2b075b37384cd91/src/effects.js#L37

It appears that this is intentional, and will just resolve to `13` in normal usage. However, with jquery-migrate loaded, that line causes a deprecation warning to be incorrectly printed.

One could argue that this should only happen on particularly old browsers, but the `if` statement above also causes the `setTimeout` path to run when the document is hidden (if the browser is minimised, for instance). This seems like a perfectly legitimate use-case.

It's not entirely clear whether this is a bug in jQuery (where `jQuery.fx.interval` claims to have been deprecated) or in jquery-migrate (where the warning is being printed despite the fact that the usage is in core jQuery code). I'd be happy to work on a pull request for whichever people think is the culprit.

dmethvin:
Fixes #292 

This seems like the easiest way to avoid the spurious error. There's no change to the tests because the current tests fire as they did before, the document isn't hidden. The test is already skipped for IE9 but if it wasn't then `document.hidden` would be falsy there anyway.

In the latest spec it looks like `document.hidden` has been [somewhat obsoleted](https://www.w3.org/TR/page-visibility/#hidden-attribute) but is still supported.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -15,11 +15,25 @@ jQuery.Tween.prototype.run = function( ) {
 	oldTweenRun.apply( this, arguments );
 };
 
-jQuery.fx.interval = jQuery.fx.interval || 13;
+var intervalValue = jQuery.fx.interval || 13,
+	intervalMsg = "jQuery.fx.interval is deprecated";
 
 // Support: IE9, Android <=4.4
 // Avoid false positives on browsers that lack rAF
+// Don't warn if document is hidden, jQuery uses setTimeout (#292)
 if ( window.requestAnimationFrame ) {
-	migrateWarnProp( jQuery.fx, "interval", jQuery.fx.interval,
-		"jQuery.fx.interval is deprecated" );
+	Object.defineProperty( jQuery.fx, "interval", {
+		configurable: true,
+		enumerable: true,
+		get: function() {
+			if ( !window.document.hidden ) {
+				migrateWarn( intervalMsg );
+			}
+			return intervalValue;
+		},
+		set: function( newValue ) {
+			migrateWarn( intervalMsg );
+			intervalValue = newValue;
+		}
+	} );
 }


mgol:
The spec mentions `visibilityState` as a replacement. It has a [pretty good browser support](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState), but, alas, it starts at IE 10 so we'd need a fallback anyway.

On the other hand, I'd rather avoid us preventing the `document.hidden` removal because we used it not as a fallback... But if we changed that we'd need to change Core as well (cc @markelog, what do you think?).

mgol:
Wait, `document.hidden` is only supported in IE 10+ as well: https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden. Perhaps we should just switch to `document.visibilityState` everywhere?

dmethvin:
We could switch I suppose. Most likely `document.hidden` will stay around forever. I'll create a ticket in core, Migrate should follow whatever we decide to do but I'll land this in the meantime.

markelog:
Mm, my understanding that if document is hidden we still use this property, is there intention to replace such logic?

mgol:
@markelog The issue is `document.hidden` is deprecated in favor of `document.visibilityState` and we should avoid deprecated APIs, especially when they have a replacement. Unfortunately, it seems IE doesn't support the new API.

markelog:
> @markelog The issue is document.hidden is deprecated in favor of document.visibilityState and we should avoid deprecated APIs, especially when they have a replacement. Unfortunately, it seems IE doesn't support the new API.

That's not what I meant :).

I am talking about jQuery.interval, with that prism my previous comment should be a bit more clear I suppose :)

markelog:
Okay, I see your point here – https://github.com/jquery/jquery-migrate/issues/292#issuecomment-371629835

I think we might want to show the warning only when this property is set – not on the getter

mgol:
@markelog That's one solution but a warning for the getter makes sense as well as we don't want people to rely on this value.

markelog:
We used as a getter, it still has fully legitimated use, but one might argue that should not be changed, that's how I see it.

But of course you can do whatever you want here :)

markelog:
> Our helpful friends at StackOverflow (😭) often seem to think it will fix some problem or another, although most involve assigning some horribly large or small value.

That's an explanation to keep the setter? Well, whatever, forget it, I don't want to argue about it

dmethvin:
@markelog not arguing at all, just letting you know that I wanted to keep the getter. This PR will eliminate the false positive for now and we can change core later.

dmethvin:
Landed at 758a1382cf9b0ec92aeb1c53d67c80dfb100e00b

PR #317, opened 2019-04-08
Build: Run tests on jQuery 3.x-git, accept browserSets & timeout params

mgol:
I've tested it works on Jenkins+TestSwarm already.

### DIFF
--- a/test/testinit.js
+++ b/test/testinit.js
@@ -139,7 +139,7 @@ TestManager = {
 TestManager.init( {
 	"jquery": {
 		urlTag: "jquery",
-		choices: "dev,min,git,3.3.1,3.2.1,3.1.1,3.0.0"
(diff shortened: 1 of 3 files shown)


test/testinit.js
```diff
@@ -139,7 +139,7 @@ TestManager = {
 TestManager.init( {
 	"jquery": {
 		urlTag: "jquery",
-		choices: "dev,min,git,3.3.1,3.2.1,3.1.1,3.0.0"
+		choices: "dev,min,git,3.x-git,3.3.1,3.2.1,3.1.1,3.0.0"
```
mgol: I left the `git` choice here but it's no longer run against `master` on TestSwarm. I assume it's fine for now, until we start making Migrate compatible with what Core 4.0 brings?

dmethvin: I think this is fine for now.

PR #319, opened 2019-05-09
Build: Support jQuery 3.4.1 and use 3.x for grunt test

dmethvin:
Since `jquery-git.js` already has some 4.x breaking changes we need to avoid using it.

I'm seeing one test failure locally on `grunt test` related to `jQuery.event.props.concat` but it's happening on 3.4.0 as well. I don't see any failures on real browsers though.

The most recent Jenkins job failed as well but I don't see anything in the output to explain it: http://jenkins.jquery.com/job/jQuery%20Migrate/412/console

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -30,7 +30,7 @@ module.exports = function( grunt ) {
 			jquery: [
 				"dev+3.x-git",
 				"min+3.x-git.min",
-				"dev+3.4.0",
+				"dev+3.4.1",
 				"dev+3.3.1",
 				"dev+3.2.1",
 				"dev+3.1.1",
@@ -116,7 +116,7 @@ module.exports = function( grunt ) {
 				},
 				frameworks: [ "qunit" ],
 				files: [
-					"https://code.jquery.com/jquery-git.min.js",
+					"https://code.jquery.com/jquery-3.x-git.min.js",
 					"dist/jquery-migrate.min.js",
 					"src/compareVersions.js",
 
--- a/test/testinit.js
+++ b/test/testinit.js
@@ -139,7 +139,7 @@ TestManager = {
 TestManager.init( {
 	"jquery": {
 		urlTag: "jquery",
-		choices: "dev,min,git,3.x-git,3.4.0,3.3.1,3.2.1,3.1.1,3.0.0"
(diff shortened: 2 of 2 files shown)


Gruntfile.js
```diff
@@ -116,7 +116,7 @@ module.exports = function( grunt ) {
 				},
 				frameworks: [ "qunit" ],
 				files: [
-					"https://code.jquery.com/jquery-git.min.js",
+					"https://code.jquery.com/jquery-3.4.1.min.js",
```
mgol: A replacement preserving the original spirit would be:
https://code.jquery.com/jquery-3.x-git.min.js
We already use it in TestSwarm, see:
http://swarm.jquery.org/job/9162

dmethvin: Yeah, I had been playing with the versions to make sure it wasn't something else and forgot to push the 3.x-git version. Updated now.

dmethvin:
Thanks for restarting the jobs!

mgol:
It's weird but I can reproduce the issue via Karma (`npm test`) but not via manually opening the QUnit page, even if I choose the same jQuery + Migrate set.

dmethvin:
I'm trying to figure this one out. It could occur, for example, if some code outside jQuery or Migrate completely overwrites `jQuery.event.props` which would wipe out the `concat` override. What makes it more complicated is that it pulls in jQuery Mobile, specifically because that was the *only* piece of code that I knew used this hack. I'm not even sure if it makes sense to support jQuery Mobile's nonsense anymore, users can always just use 2.x or something. I do see the message appearing on the console if I run the test standalone though.

ecadius:
There are also security issues in 2.x https://www.cvedetails.com/vulnerability-list.php?vendor_id=6538
How can we use this version?

mgol:
@ecadius As far as I know, jQuery Mobile itself still doesn't support jQuery 3.0 & later officially, we can't do that work on the jQuery+Migrate side ourselves.

dmethvin:
I merged this in 433656020ff77f9bd26eedcdae1ee24c78061689 and in the process had to do a bunch of other housekeeping. On Windows the existing deps wouldn't build properly.

mgol:
> On Windows the existing deps wouldn't build properly.

Not only on Windows!

PR #320, opened 2019-05-27
Update copyright and remove AUTHORS

dmethvin:
This isn't critical but I figured it might be worth doing before the next release. I'm also proposing that we remove the AUTHORS file which is a bit of a maintenance headache. The full list of contributors is available via the git log and [GitHub contributor page](https://github.com/jquery/jquery-migrate/graphs/contributors). It looks like we're missing some right now, plus it doesn't acknowledge contributions from issue reporters.


mgol:
I think the motivation for the `AUTHORS` file was that the GitHub contributors page only shows people with GitHub accounts. Some older contributors to UI (& maybe Core) are not attached to a GitHub account so their contributions wouldn't be highlighted.

mgol:
The AUTHORS file is updated via `grunt authors`; running this task is on the jQuery Core [Release Process](https://github.com/jquery/jquery/wiki/Release-process) list. jquery-release is supposed to check that (not sure if that happens for Migrate).

The task is defined in `grunt-git-authors` so if Migrate was getting rid of that this package should go away as well.

dmethvin:
Yeah I can see that perhaps for a really old project, but we don't have that situation with Migrate. A shorter manual checklist increases the odds that it will be done right! I've removed the Grunt plugin so that cleans everything up I think.

dmethvin:
Duh, fixed.

PR #321, opened 2019-05-29
Build: Process the map file correctly

dmethvin:
Since the minified file includes a map reference it was about as easy
to fix the reference as it was to remove it.

I thought about editing the ESLint config so I could clean up this file and use modern JS but decided it can wait until later. Also, our style guide sucks. 😸

### DIFF
--- a/build/release.js
+++ b/build/release.js
@@ -286,7 +286,7 @@ function fixMinRef( oldText ) {
 	// Remove the ref for now rather than try to fix it
 	var newText = oldText.replace( mapRef, "" );
 	if ( oldText === newText ) {
-		throw "fixMinRef: Unable to find the sourceMappingURL";
+		throw Error( "fixMinRef: Unable to find the sourceMappingURL" );
 	}
 	return newText;
 }
--- a/build/release.js
+++ b/build/release.js
@@ -295,7 +295,7 @@ function fixMapRef( oldText, newFile ) {
 	var mapJSON = JSON.parse( oldText );
 	var sources = mapJSON.sources;
 	if ( sources.join() !== "../src/migratemute.js,jquery-migrate.js" ) {
-		throw "fixMapRef: Unexpected sources entry: " + sources;
+		throw Error( "fixMapRef: Unexpected sources entry: " + sources );
 	}
 
 	// This file isn't published, not sure the best way to deal with that
--- a/build/release.js
+++ b/build/release.js
@@ -14,7 +14,6 @@ var fs = require( "fs" ),
 
 var releaseVersion,
 	nextVersion,
-	finalFiles,
 	isBeta,
 	pkg,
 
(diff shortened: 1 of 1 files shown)


mgol:
One major reason we are stripping the source mapping comment from Core is that our files have different names on different CDNs and that requires changing the referenced name. That, in turn, makes Subresource Integrity hash different which we wanted to avoid.

Because of that, I feel it'd be best to leave out the comment here as well.

dmethvin:
After a night's sleep, I agree. The map isn't going to work quite right anyway since it references the migratemute.js file which isn't published on the CDN. 

How about if we remove the `sourceMappingURL` comment but publish the map? That way someone can associate the map manually.

mgol:
> How about if we remove the `sourceMappingURL` comment but publish the map? That way someone can associate the map manually.

Yes, that sounds like the best way. Plus, we're doing that in Core already. :)

dmethvin:
Okay, this seems to be working fine now. I'm also removing Windows line endings as a precaution. I think this is everything before the release of 3.1.0 and I've tagged all the issues that went into it.

build/release.js
```diff
@@ -272,11 +280,28 @@ function writeJsonSync( fname, json ) {
...
+
+function fixMapRef( oldText, newFile ) {
+	var mapJSON = JSON.parse( oldText );
+	var sources = mapJSON.sources;
+	if ( sources.join() !== "../src/migratemute.js,jquery-migrate.js" ) {
+		throw "fixMapRef: Unexpected sources entry: " + sources;
 	}
+
+	// This file isn't published, not sure the best way to deal with that
+	sources[ 0 ] = "migratemute.js";
```
mgol: Why do we build it if we don't publish it?

build/release.js
```diff
@@ -272,11 +280,28 @@ function writeJsonSync( fname, json ) {
...
-	status( "Copying " + oldFile + " to " + newFile );
-	if ( !dryrun ) {
-		fs.writeFileSync( newFile, fs.readFileSync( oldFile, "utf8" ) );
+function fixMinRef( oldText ) {
+	var mapRef = new RegExp( "^//# sourceMappingURL=jquery-migrate.min.map\\n?", "m" );
+
+	// Remove the ref for now rather than try to fix it
+	var newText = oldText.replace( mapRef, "" );
+	if ( oldText === newText ) {
+		throw "fixMinRef: Unable to find the sourceMappingURL";
```
mgol: Let's throw errors, not strings so that stack traces are preserved:
```suggestion
		throw Error( "fixMinRef: Unable to find the sourceMappingURL" );
```

build/release.js
```diff
@@ -272,11 +280,28 @@ function writeJsonSync( fname, json ) {
...
+	}
+	return newText;
+}
+
+function fixMapRef( oldText, newFile ) {
+	var mapJSON = JSON.parse( oldText );
+	var sources = mapJSON.sources;
+	if ( sources.join() !== "../src/migratemute.js,jquery-migrate.js" ) {
+		throw "fixMapRef: Unexpected sources entry: " + sources;
```
mgol:
```suggestion
		throw Error( "fixMapRef: Unexpected sources entry: " + sources );
```

build/release.js
dmethvin: Right now the way we turn off warnings in the minified file is to [include `migratemute.js` in front of the standard file](https://github.com/jquery/jquery-migrate/blob/master/Gruntfile.js#L79-L80) and run uglify on that. As a result it makes a map that mentions both files. However we only publish the standard unminified file, not `migratemute.js`. I don't see any direct uglify options to deal with this kind of thing, which isn't surprising TBH because it's a pretty strange way to make a map. I've got some ideas on how to improve it but am thinking it might be best to just publish first since this map isn't even referenced in the minified file at the moment.

dmethvin:
Thanks for the sanity check @mgol, not sure what I was thinking when I `threw` those strings!

On the map file itself, I think the solution is to concat `migratemute.js` onto the first line of the unminified file and run uglify on that. The line numbers should stay consistent that way. I just want some time to play around with that.

PR #322, opened 2019-06-09
Core: Get pre-4.0 compat by implementing removed functions

dmethvin:
If possible I'd like to avoid having Migrate 3 break when used on jQuery 4. That was inevitable with jQuery 1/2 and Migrate 1 because we changed the behaviors of several APIs which made it impossible to sanely build Migrate 3 on Migrate 1. I think we can do better here.

This PR adds the jQuery core implementation and unit tests for `jQuery.type`, `jQuery.isNumeric`, and `jQuery.nodeName`. It also removes the Migrate internal references to those APIs. For now at least, this runs against jQuery pre-4.x master.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -84,18 +84,20 @@ migrateWarnProp( jQuery.expr, ":", jQuery.expr.pseudos,
 
 // Prior to jQuery 3.2 there were internal refs so we don't warn there
 if ( jQueryVersionSince( "3.2.0" ) ) {
-	migrateWarnFunc( jQuery, "nodeName", jQuery.nodeName,
+	migrateWarnFunc( jQuery, "nodeName", function( elem, name ) {
+		return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
+	},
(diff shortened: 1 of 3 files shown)


mgol:
> If possible I'd like to avoid having Migrate 3 break when used on jQuery 4.

Will Migrate need to include Sizzle to support positional selectors via overriding `jQuery.find`? Maybe that's not a bad idea.

We also have an event system overhaul in plans; it remains to be seen how backwards-compatible that will be.

dmethvin:
> Will Migrate need to include Sizzle to support positional selectors via overriding jQuery.find? 

That seems like a good idea. We could at least do the simple cases of selector strings that end in positional selectors for example. If it's not much harder to do the embedded cases by splitting them I'm good with that too. In some ways I'd prefer to just warn on them rather than giving people a reliable crutch that encourages them to use Migrate forever.

dmethvin:
Merged as d86d4e1798922251d6a344416f9bda6e7b94ead6

PR #328, opened 2020-03-18
Manipulation: Allow to restore the legacy jQuery.htmlPrefilter

mgol:
This goes in tandem with https://github.com/jquery/jquery.com/pull/196.


mgol:
This goes in tandem with https://github.com/jquery/jquery.com/pull/196.

src/manipulation.js
```diff
@@ -0,0 +1,7 @@
+var rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\0>\x20\t\r\n\f]*)[^>]*)\/>/gi;
+
+jQuery.restoreLegacyInsecureHtmlPrefilter = function() {
```
mgol: Optionally, to make people really think before invoking that, we can do:
```suggestion
jQuery.UNSAFE_restoreLegacyInsecureHtmlPrefilter = function() {
```

dmethvin: Maybe `UNSAFE_restoreHtmlPrefilter`?

mgol: @dmethvin That seems to suggest to me that we don't have `htmlPrefilter` now but we do, just with a different value.

dmethvin: How about just `jQuery.UNSAFE_restoreLegacyHtmlPrefilter` since we have UNSAFE in the name now.

mgol: PR updated.

mgol:
@dmethvin What renaming? I've renamed it a few hours before your PR approval...

dmethvin:
Strange, I must have commented on the earlier commit. In any case, this looks ready to go.

Issue #325, opened 2019-08-26
Warn & fill jQuery.trim


mgol:
`jQuery.trim` has been deprecated in https://github.com/jquery/jquery/issues/4363, slated for jQuery 4.0.0. Migrate should warn against it & fill it.

Considering that the deprecation will only happen in 4.0, should Migrate perhaps for now just fill (for forward compat) & warn later? Or do we want to wait with the whole issue? cc @dmethvin

mgol:
API issue: https://github.com/jquery/api.jquery.com/issues/1142

mgol:
PR: #331

PR #331, opened 2020-03-28
Core: Warn & fill jQuery.trim

mgol:
Fixes gh-325

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -2,7 +2,11 @@
 var oldInit = jQuery.fn.init,
 	oldFind = jQuery.find,
 	rattrHashTest = /\[(\s*[-\w]+\s*)([~|^$*]?=)\s*([-\w#]*?#[-\w#]*)\s*\]/,
-	rattrHashGlob = /\[(\s*[-\w]+\s*)([~|^$*]?=)\s*([-\w#]*?#[-\w#]*)\s*\]/g;
+	rattrHashGlob = /\[(\s*[-\w]+\s*)([~|^$*]?=)\s*([-\w#]*?#[-\w#]*)\s*\]/g,
+
+	// Support: Android <=4.0 only
+	// Make sure we trim BOM and NBSP
+	rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
 
 jQuery.fn.init = function( arg1 ) {
 	var args = Array.prototype.slice.call( arguments );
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -60,15 +61,15 @@ for ( findProp in oldFind ) {
 }
 
 // The number of elements contained in the matched element set
-jQuery.fn.size = function() {
-	migrateWarn( "jQuery.fn.size() is deprecated and removed; use the .length property" );
+migrateWarnFunc( jQuery.fn, "size", function() {
```
mgol: This is not related but I thought I might as well use the helper we have

test/core.js
```diff
@@ -370,15 +370,63 @@ QUnit.test( "jQuery.holdReady (warn only)", function( assert ) {
 	} );
 } );
 
+QUnit[ jQueryVersionSince( "3.1.1" ) ? "test" : "skip" ]( "jQuery.trim", function( assert ) {
+	assert.expect( 26 );
+
+	var nbsp = String.fromCharCode( 160 );
+
+	expectWarning( assert, "jQuery.trim", 1, function() {
```
mgol: I split it like that as Migrate now only warns once against any API used twice... Maybe we should have a test helper here so that we can track how many times a warning is detected even if it's skipped?

src/core.js
```diff
@@ -82,6 +83,14 @@ migrateWarnProp( jQuery.expr, "filters", jQuery.expr.pseudos,
 migrateWarnProp( jQuery.expr, ":", jQuery.expr.pseudos,
 	"jQuery.expr[':'] is deprecated; use jQuery.expr.pseudos" );
 
+// Prior to jQuery 3.1.1 there were internal refs so we don't warn there
+if ( jQueryVersionSince( "3.1.1" ) ) {
+	migrateWarnFunc( jQuery, "trim", function( text ) {
+		return text == null ? "" : trim.call( text );
```
dmethvin: The [3.x code](https://github.com/jquery/jquery/blob/3.4.1/src/core.js#L271) uses `(text + "")` to ensure the argument is stringified in case it's something unusual. Seems like we'd want to do that here too in order for any edge cases to work.

test/core.js
dmethvin: I'm fine with separate tests really. In real code once you know you're looking for `.trim` you can just search for that.

src/core.js
dmethvin: Oh duh, we don't need it here!

mgol: Hmm, maybe I should use the 3.x version to keep compat with Android 4.0... After all, people can use jQuery 3.x with Migrate 3.x to prepare for a future jQuery 4.x upgrade even if they’re not ready to drop browsers not supported by jQuery 4.x yet.

dmethvin: Yeah, I agree with that. We have the code there and don't care about bytes in Migrate so we might as well use the same code.

mgol: PR updated.

mgol:
@dmethvin I forgot about an entry in warnings.md, oops! Added, please have another look.

test/core.js
mgol: @dmethvin I submitted #332 acting on the idea proposed in my comment above.

PR #332, opened 2020-04-01
Core:Tests: Add jQuery.migrateDeduplicateWarnings; don't dedupe in tests

mgol:
This commit introduces a new API, `jQuery.migrateDeduplicateWarnings`, set to
`true` by default but set to `false` in tests. This allows a more reliable
testing of bigger code blocks wrapped in a single `expectWarning` helper.

The commit also leverages this new behavior to simplify `jQuery.trim` tests.

Ref gh-331

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -24,6 +24,27 @@ QUnit.test( "compareVersions and jQueryVersionSince", function( assert ) {
 	assert.equal( jQueryVersionSince( jQuery.fn.jquery ), true, "since - equal" );
 } );
 
+QUnit.test( "jQuery.migrateDeduplicateWarnings", function( assert ) {
+	assert.expect( 3 );
+
+	var origValue = jQuery.migrateDeduplicateWarnings;
+	assert.strictEqual( origValue, true, "true by default" );
+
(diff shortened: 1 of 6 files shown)


test/core.js
```diff
@@ -364,56 +385,32 @@ QUnit.test( "jQuery.expr.pseudos aliases", function( assert ) {
...
 } );
 
 QUnit[ jQueryVersionSince( "3.1.1" ) ? "test" : "skip" ]( "jQuery.trim", function( assert ) {
-	assert.expect( 26 );
+	assert.expect( 14 );
 
 	var nbsp = String.fromCharCode( 160 );
 
-	expectWarning( assert, "jQuery.trim", 1, function() {
+	expectWarning( assert, "jQuery.trim", 13, function() {
```
dmethvin, line 399: Well that shortened everything quite a bit!

mgol:
@dmethvin I like the wording, PR updated.

I initially thought about it for test usage only but it might be useful even for Migrate consumption: if someone is working on resolving Migrate warnings, disabling deduplication together with filtering in the DevTools console would show them all violations all in one place, making it potentially easier to fix them all in one go.

PR #334, opened 2020-04-02
Offset: fix wrong origin value

### LINKED ISSUE #330: Why `offset` is restored to return `{ left: 0, top: 0}` in 3.1.0?
jQuery 2.2.4 is already returning `undefined` for an empty collection. Why jquery-migrate 3.1.0 is restoring it to return `{ left: 0, top: 0 }`?

https://github.com/jquery/jquery-migrate/blob/3.1.0/src/offset.js

panlina:
Fixes #330 .

`offset` should return `undefined` instead of `{top:0,left:0}` for non DOM element.

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -4,17 +4,17 @@ var oldOffset = jQuery.fn.offset;
 jQuery.fn.offset = function() {
 	var docElem,
 		elem = this[ 0 ],
-		origin = { top: 0, left: 0 };
+		bogus = { top: 0, left: 0 };
 
 	if ( !elem || !elem.nodeType ) {
 		migrateWarn( "jQuery.fn.offset() requires a valid DOM element" );
-		return origin;
+		return undefined;
 	}
 
 	docElem = ( elem.ownerDocument || window.document ).documentElement;
 	if ( !jQuery.contains( docElem, elem ) ) {
 		migrateWarn( "jQuery.fn.offset() requires an element connected to a document" );
-		return origin;
+		return bogus;
 	}
 
 	return oldOffset.apply( this, arguments );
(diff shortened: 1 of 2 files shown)


panlina:
When will this be merged?

mgol:
I'd like @dmethvin to do a review as well but I'd like to get it in a Migrate 3.2.0 release that is likely to be available within a week or two.

src/offset.js
```diff
@@ -4,17 +4,17 @@ var oldOffset = jQuery.fn.offset;
 jQuery.fn.offset = function() {
 	var docElem,
 		elem = this[ 0 ],
-		origin = { top: 0, left: 0 };
+		bogus = { top: 0, left: 0 };
 
 	if ( !elem || !elem.nodeType ) {
 		migrateWarn( "jQuery.fn.offset() requires a valid DOM element" );
-		return origin;
```
dmethvin, line 11: We only want to return `undefined` for the `!elem` case. Some code including older older versions of jQuery UI would pass `window` and expect the top/left of 0. https://github.com/jquery/jquery/issues/3249

panlina: Does it expect `{left:0,top:0}` for `window`? Why both `1.12.4` and `2.2.4` return `undefined` as I tested?

dmethvin: It sure does, thanks for catching that! I don't have time to look at it right now but I must have been thinking of some other edge case with `.offset()`? Looks like that ticket is specifically about calling a method on `window`, not returning the bogus value.

mgol: @dmethvin Since the issue was about a breaking change in jQuery 3.0.0, it seems we shouldn't care about it here as Migrate 3.x is meant to aid with upgrading to the future jQuery 4.0. Am I missing anything?

PR #336, opened 2020-04-10
Build: Test on jQuery 4.x separately to jQuery 3.x

mgol:
Having separate test runs for those jQuery lines is necessary as they support
different browsers. Android 4.0 fails hard when run against jQuery master code.

A git.min version was also added to aid manual testing.

I've already modified Jenkins jobs to adhere to what this PR does. I renamed the existing "jQuery Migrate" to "jQuery Migrate - 4.x core" & created a new "jQuery Migrate - 3.x core" job. Both jobs will now fire for every new Migrate commit.

I run a test run on my fork and both jobs passed:
jQuery Migrate - 4.x core: http://swarm.jquery.org/job/9632
jQuery Migrate - 3.x core: http://swarm.jquery.org/job/9633

### DIFF
--- a/test/testinit.js
+++ b/test/testinit.js
@@ -157,7 +157,7 @@ TestManager = {
 TestManager.init( {
 	"jquery": {
 		urlTag: "jquery",
-		choices: "dev,min,git,3.x-git,3.x-git.min,3.4.1,3.3.1,3.2.1,3.1.1,3.0.0"
+		choices: "dev,min,git,git.min,3.x-git,3.x-git.min,3.5.0,3.4.1,3.3.1,3.2.1,3.1.1,3.0.0"
 	},
 	"jquery-migrate": {
 		urlTag: "plugin",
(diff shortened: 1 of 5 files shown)


test/testinit.js
```diff
@@ -157,7 +157,7 @@ TestManager = {
 TestManager.init( {
 	"jquery": {
 		urlTag: "jquery",
-		choices: "dev,min,git,3.x-git,3.x-git.min,3.4.1,3.3.1,3.2.1,3.1.1,3.0.0"
+		choices: "dev,min,git,git.min,3.x-git,3.x-git.min,3.4.1,3.3.1,3.2.1,3.1.1,3.0.0"
```
dmethvin, line 160: Do you want to add `3.5.0` while you are here?

mgol: Good point, will do.

mgol: Done.

PR #337, opened 2020-04-10
CSS: Warn and fill .css(name, Number)

### LINKED ISSUE #296: Warn about jQuery.cssNumber
See https://github.com/jquery/jquery/issues/4009

Detect assignments to `jQuery.cssNumber` and recommend that it only be done if that property exists; note that it's going away in 4.0.

Recommend that people do not pass raw numbers to `.css()` and instead always pass strings and append `"px"` when appropriate.  Point out `line-height` as a troublesome case.

mgol:
Fixes #296

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -83,14 +83,10 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
 	} );
 }
 
+// Create a dummy jQuery.cssNumber if missing. It won't be used by jQuery but
+// it will prevent code adding new keys to it unconditionally from crashing.
 if ( !jQuery.cssNumber ) {
 	jQuery.cssNumber = {};
-}
-
-// jQuery 3.x uses jQuery.cssNumber internally so we can't warn on access there.
-if ( jQueryVersionSince( "4.0.0" ) ) {
-	migrateWarnProp( jQuery, "cssNumber", jQuery.cssNumber,
-		"jQuery.cssNumber is deprecated" );
 }
 
 function isAutoPx( prop ) {
--- a/src/css.js
+++ b/src/css.js
@@ -52,3 +52,25 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
 		}
 	} );
 }
+
+if ( !jQuery.cssNumber ) {
+	jQuery.cssNumber = {};
+}
+migrateWarnProp( jQuery, "cssNumber", jQuery.cssNumber,
+	"jQuery.cssNumber is deprecated" );
+
+var oldFnCss = jQuery.fn.css;
+
+jQuery.fn.css = function( name, value ) {
+	var origThis = this;
+	if ( typeof name !== "string" ) {
+		jQuery.each( name, function( n, v ) {
+			jQuery.fn.css.call( origThis, n, v );
+		} );
+	}
+	if ( typeof value === "number" ) {
+		migrateWarn( "Use of number-typed values is deprecated in jQuery.fn.css" );
+	}
+
+	return oldFnCss.apply( this, arguments );
+};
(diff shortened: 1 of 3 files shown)


mgol:
@dmethvin The first commit is yours, my modifications in the second one.

src/css.js
```diff
@@ -52,3 +82,38 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
...
+
+if ( !jQuery.cssNumber ) {
+	jQuery.cssNumber = {};
+}
+
+// jQuery 3.x uses jQuery.cssNumber internally so we can't warn on access there.
+if ( jQueryVersionSince( "4.0.0" ) ) {
+	migrateWarnProp( jQuery, "cssNumber", jQuery.cssNumber,
+		"jQuery.cssNumber is deprecated" );
+}
```
mgol, line 94: @dmethvin I wonder if this warning is useful even for jQuery 4.x. People may still need conditional code adding more props to that object and then they'd hit the warning even if they guard access with an `if` statement. It seems impossible to warn only on invalid usage. Should we remove it?

src/css.js
```diff
@@ -52,3 +82,38 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
...
+
+function isAutoPx( prop ) {
+
+	// The first test is used to ensure that:
+	// 1. The prop starts with a lowercase letter (as we uppercase it for the second regex).
+	// 2. The prop is not empty.
+	return ralphaStart.test( prop ) &&
+		rautoPx.test( prop[ 0 ].toUpperCase() + prop.slice( 1 ) );
+}
```
mgol, line 103: Copied from https://github.com/jquery/jquery/blob/master/src/css/isAutoPx.js.

src/css.js
```diff
@@ -1,5 +1,35 @@
...
+	//        |         |  / Margin  \  |   | | Right  | |    |
+	//        |---------+-|           |-+---+-| Bottom |-+----|
+	//        |            \ Padding /         \ Left /       |
+	// BEGIN -|                                               |
+	//        |                /---------\                    |
+	//        |               |           |                   |
+	//        |               |  / Min \  |    / Width  \     |
+	//         \--------------+-|       |-+---|          |---/
+	//                           \ Max /       \ Height /
+	rautoPx = /^(?:Border(?:Top|Right|Bottom|Left)?(?:Width|)|(?:Margin|Padding)?(?:Top|Right|Bottom|Left)?|(?:Min|Max)?(?:Width|Height))$/;
```
mgol, line 26: Copied from https://github.com/jquery/jquery/blob/master/src/css/isAutoPx.js.

test/css.js
```diff
@@ -41,3 +41,95 @@ QUnit[ ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) ? "test
...
+		"border",
+		"borderWidth",
+		"borderTop",
+		"borderTopWidth",
+		"borderRight",
+		"borderRightWidth",
+		"borderBottom",
+		"borderBottomWidth",
+		"borderLeft",
+		"borderLeftWidth"
```
mgol, line 78: These are all the possible strings that report `true` when passed to `isAutoPx()`.

src/css.js
dmethvin: I agree, let's remove it.

mgol:
@dmethvin I removed warning on `jQuery.cssNumber` & added one comment & test. All in the last commit. PTAL.

mgol:
@dmethvin This is a new warning so we probably need a new minor release, not a patch one. So the next one would be `3.3.0`. Does that sound good? If yes then we need to change the `3.2.1` milestone to `3.3.0`.

dmethvin:
Agreed, I've renamed the next milestone.

PR #338, opened 2020-04-11
Core: detect incorrectly nested elements

dmethvin:
Basic approach: If our old `jQuery.htmlPrefilter` would have changed the string _and_ the resulting actual HTML string is different, give a warning.

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -4,18 +4,22 @@ var rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\
 		var doc = window.document.implementation.createHTMLDocument( "" );
 		doc.body.innerHTML = html;
 		return doc.body && doc.body.innerHTML;
+	},
+	warnIfChanged = function( html ) {
+		var changed = html.replace( rxhtmlTag, "<$1></$2>" );
+		if ( changed !== html && makeMarkup( html ) !== makeMarkup( changed ) ) {
+			migrateWarn( "HTML tags must be properly nested and closed: " + html );
+		}
 	};
 
 jQuery.UNSAFE_restoreLegacyHtmlPrefilter = function() {
 	jQuery.htmlPrefilter = function( html ) {
+		warnIfChanged( html );
 		return html.replace( rxhtmlTag, "<$1></$2>" );
 	};
 };
 
 jQuery.htmlPrefilter = function( html ) {
-	var changed = html.replace( rxhtmlTag, "<$1></$2>" );
-	if ( changed !== html && makeMarkup( html ) !== makeMarkup( changed ) ) {
-		migrateWarn( "HTML tags must be properly nested and closed: " + html );
-	}
+	warnIfChanged( html );
 	return origHtmlPrefilter( html );
 };
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -1,7 +1,21 @@
-var rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\0>\x20\t\r\n\f]*)[^>]*)\/>/gi;
+var rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\0>\x20\t\r\n\f]*)[^>]*)\/>/gi,
+	origHtmlPrefilter = jQuery.htmlPrefilter,
+	makeMarkup = function( html ) {
+		var doc = window.document.implementation.createHTMLDocument( "" );
+		doc.body.innerHTML = html;
+		return doc.body && doc.body.innerHTML;
+	};
 
 jQuery.UNSAFE_restoreLegacyHtmlPrefilter = function() {
 	jQuery.htmlPrefilter = function( html ) {
 		return html.replace( rxhtmlTag, "<$1></$2>" );
 	};
+};
+
+jQuery.htmlPrefilter = function( html ) {
+	var changed = html.replace( rxhtmlTag, "<$1></$2>" );
+	if ( changed !== html && makeMarkup( html ) !== makeMarkup( changed ) ) {
+		migrateWarn( "HTML tags must be properly nested and closed: " + html );
+	}
+	return origHtmlPrefilter( html );
 };
--- a/test/manipulation.js
+++ b/test/manipulation.js
@@ -2,6 +2,42 @@
 
 QUnit.module( "manipulation" );
 
+QUnit.test( "Improperly closed elements", function( assert ) {
+	assert.expect( 4 );
+
+	var oldWindowAlert = window.alert;
+	window.alert = function() {
+		assert.notOk( "Called alert with " + arguments );
+	};
+
+	expectWarning( assert, "Elements not self-closable nested wrong", 4, function() {
+		jQuery( "<div><p/><span/></div>" );
+		jQuery( "<blockquote><p class='bad' /><span/></blockquote>" );
+		jQuery( "<div />" ).append( "<div data-borked='y'><span/> <p/></div>" );
+		jQuery( "<script /><div class=afterscript/>" );
+	} );
+
+	expectNoWarning( assert, "Script doesn't run in processed strings", function() {
+		jQuery( "<script>alert( 'YOU SHOULD NOT SEE THIS ALERT' )</script>" );
+		jQuery( "<div>" ).append( "<script src='YOU-SHOULD-NOT-SEE-THIS.SCRIPT'></script>" );
+		jQuery( "<script>alert( 'YOU SHOULD NOT SEE THIS CONSOLE' )</script>" );
+	} );
+
+	expectNoWarning( assert, "Elements not self-closable but tolerable", function() {
+		jQuery( "<div class=wonky />" );
+		jQuery( "<p style='width: 2%' />" );
+		jQuery( "<p />" ).append( "<span aria-label='hello' />" );
+	} );
+
+	expectNoWarning( assert, "Bare elements", function() {
+		jQuery( "<p/>" ).append( "<strong>" );
+		jQuery( "<abbr />" );
+		jQuery( "<head />" );
+	} );
+
+	window.alert = oldWindowAlert;
+} );
+
 // Do this as iframe because there is no way to undo `jQuery.UNSAFE_restoreLegacyHtmlPrefilter()`
 TestManager.runIframeTest( "jQuery.UNSAFE_restoreLegacyHtmlPrefilter",
 	"manipulation-UNSAFE_restoreLegacyHtmlPrefilter.html",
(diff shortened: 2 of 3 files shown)


src/core.js
```diff
@@ -3,13 +3,28 @@ var oldInit = jQuery.fn.init,
...
 	rattrHashGlob = /\[(\s*[-\w]+\s*)([~|^$*]?=)\s*([-\w#]*?#[-\w#]*)\s*\]/g,
+	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\0>\x20\t\r\n\f]*)[^>]*)\/>/gi,
+	rsingleTag = ( /^<([a-z][^\/\0>:\x20\t\r\n\f]*)[\x20\t\r\n\f]*\/?>(?:<\/\1>|)$/i ),
 
 	// Support: Android <=4.0 only
 	// Make sure we trim BOM and NBSP
 	rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
 
+var makeMarkup = function( html ) {
+	var createDoc = window.document.implementation.createHTMLDocument;
```
dmethvin, line 14: Not completely sure this guard is needed yet.

dmethvin:
@mgol  This seems to work pretty well. Thoughts?

src/core.js
```diff
@@ -18,6 +33,15 @@ jQuery.fn.init = function( arg1 ) {
 		args[ 0 ] = [];
 	}
 
+	if ( typeof arg1 === "string" && arg1.trim().charAt( 0 ) === "<" ) {
+		if ( !rsingleTag.test( arg1 ) ) {
+			changed = arg1.replace( rxhtmlTag, "<$1></$2>" );
+			if ( changed !== arg1 && makeMarkup( arg1 ) !== makeMarkup( changed ) ) {
+				migrateWarn( "HTML tags must be properly nested and closed: " + arg1 );
```
mgol, line 40: I was thinking about doing it by overwriting `jQuery.htmlPrefilter` instead; then we would miss no cases & the logic would be simpler. Of course, when someone overwrites it on their end, that would remove the warnings but maybe that's acceptable? We could make it more explicit in the [3.5 upgrade guide](https://jquery.com/upgrade-guide/3.5/) that if Migrate is used, it's best to avoid manually overwriting the method?

Doing it in `jQuery.fn.init` probably catches the majority of usage but I suspect it leaves a significant portion out. Consider e.g.:
```js
jQuery( "#test" ).append( "<div/><span/>" );
```
We'd get no warning for that.

dmethvin:
Using `.htmlPrefilter()` seems like a good idea, I'll switch to that and add a few more tests.

src/core.js
```diff
@@ -22,6 +36,19 @@ jQuery.fn.init = function( arg1 ) {
 };
 jQuery.fn.init.prototype = jQuery.fn;
 
+jQuery.htmlPrefilter = function( html ) {
+	var changed;
+	if ( typeof html === "string" && html.trim().charAt( 0 ) === "<" ) {
```
dmethvin, line 41: Not sure if this guard is needed but it shouldn't hurt.

test/core.js
```diff
@@ -1,84 +1,122 @@
 
-QUnit.module( "core" );
```
dmethvin, line 2: I'm happier with this lower-space version but not sure why it happened. Did we change some config file?

test/core.js
```diff
@@ -1,84 +1,122 @@
...
+		jQuery("<blockquote><p class='bad' /><span/></blockquote>");
+		jQuery("<div />").append("<div data-borked='y'><span/> <p/></div>");
+		jQuery("<script /><div class=afterscript/>")
+	});
+
+	expectNoWarning(assert, "Script doesn't run in processed strings", function () {
+		jQuery("<script>alert('YOU SHOULD NOT SEE THIS ALERT')</script>");
+		jQuery("<div>").append("<script src='YOU-SHOULD-NOT-SEE-THIS.SCRIPT'></script>");
+		jQuery("<script>alert('YOU SHOULD NOT SEE THIS CONSOLE')</script>");
+		// jQuery core code causes this to execute, and also it's async
```
dmethvin, line 87: I wonder if we should use this approach for jQuery 4.0 to avoid running script in the initial processing.

dmethvin:
@mgol Take a look when you get a chance and give me your thoughts.

src/core.js
mgol: The original `jQuery.htmlPrefilter` in jQuery 3.4.1 & older is just:
```js
jQuery.htmlPrefilter = function( html ) {
	return html.replace( rxhtmlTag, "<$1></$2>" );
};
```
so it has no guard. Therefore, I don't think we need it here either.

test/core.js
mgol: I submitted a PR that improves our ESLint setup and should resolve the issue you had here: #339. Let's get it merged first so that we don't go back & forth on code style here.

test/core.js
mgol: This follows the native behavior, I wouldn't want to fight it too much.

test/core.js
```diff
@@ -1,84 +1,122 @@
...
-				"<a href='space#junk'>test</a>" +
-				"<link rel='good#stuff' />" +
-				"<p class='space #junk'>" +
-					"<a href='#some-anchor'>anchor2</a>" +
-					"<input value='[strange*=#stuff]' />" +
-					"<a href='#' data-id='#junk'>anchor</a>" +
-				"</p>" +
-			"</div>" ).appendTo( "#qunit-fixture" ),
+		"<div>" +
+		"<div data-selector='a[href=#main]'></div>" +
```
mgol, line 111: The previous markup was more readable due to indentation. I'd keep the old styling or add indentation in the string itself, the new way is hard to read.

IMO it'd be good to revert all unrelated style changes from this PR, merge #339 & then only add new tests. We can always improve the styling in a separate PR later when we know ESLint actually checks it.

test/core.js
mgol: #339 has landed so you can rebase now.

test/core.js
```diff
@@ -60,12 +60,48 @@ QUnit.test( "jQuery( '#' )", function( assert ) {
 	assert.expect( 2 );
 
 	expectWarning( assert, "Selector, through the jQuery constructor, nothing but hash",
-			function() {
```
dmethvin, line 63: There are some spurious diffs here from line lengths exceeding 100 chars that I had to fix.

src/core.js
```diff
@@ -25,6 +31,14 @@ jQuery.fn.init = function( arg1 ) {
 };
 jQuery.fn.init.prototype = jQuery.fn;
 
+jQuery.htmlPrefilter = function( html ) {
+	var changed = html.replace( rxhtmlTag, "<$1></$2>" );
+	if ( changed !== html && makeMarkup( html ) !== makeMarkup( changed ) ) {
+		migrateWarn( "HTML tags must be properly nested and closed: " + html );
+	}
+	return html;
+};
```
mgol, line 40: This will make `jQuery.htmlPrefilter` in jQuery <3.5 to behave like an identity function which we should avoid IMO. I'd keep the warning mechanism as is as it will prepare users for a 3.5 upgrade but instead of returning `html` I'd return the result of applying the original cached `htmlPrefilter`:
```suggestion
var origHtmlPrefilter = jQuery.htmlPrefilter;
jQuery.htmlPrefilter = function( html ) {
	var changed = html.replace( rxhtmlTag, "<$1></$2>" );
	if ( changed !== html && makeMarkup( html ) !== makeMarkup( changed ) ) {
		migrateWarn( "HTML tags must be properly nested and closed: " + html );
	}
	return origHtmlPrefilter( html );
};
```

test/core.js
```diff
@@ -60,12 +60,48 @@ QUnit.test( "jQuery( '#' )", function( assert ) {
...
+		jQuery( "<blockquote><p class='bad' /><span/></blockquote>" );
+		jQuery( "<div />" ).append( "<div data-borked='y'><span/> <p/></div>" );
+		jQuery( "<script /><div class=afterscript/>" );
+	} );
+
+	expectNoWarning( assert, "Script doesn't run in processed strings", function() {
+		jQuery( "<script>alert( 'YOU SHOULD NOT SEE THIS ALERT' )</script>" );
+		jQuery( "<div>" ).append( "<script src='YOU-SHOULD-NOT-SEE-THIS.SCRIPT'></script>" );
+		jQuery( "<script>alert( 'YOU SHOULD NOT SEE THIS CONSOLE' )</script>" );
+	} );
```
mgol, line 88: This doesn't hurt but out of curiosity: is this related to the tag self-closing behavior?

src/core.js
```diff
@@ -25,6 +31,14 @@ jQuery.fn.init = function( arg1 ) {
 };
 jQuery.fn.init.prototype = jQuery.fn;
 
+jQuery.htmlPrefilter = function( html ) {
```
mgol, line 34: I think this code should lie in `src/manipulation.js` as it was in jQuery. In that case you won't need a separate `rxhtmlTag` definition as it's already defined there.

test/core.js
dmethvin: Mainly I just wanted to make sure we weren't introducing any script issues with this.

src/core.js
dmethvin: I'll move it. I'd actually forgotten we had it there.

src/manipulation.js
```diff
@@ -1,7 +1,21 @@
...
+		return doc.body && doc.body.innerHTML;
+	};
 
 jQuery.UNSAFE_restoreLegacyHtmlPrefilter = function() {
 	jQuery.htmlPrefilter = function( html ) {
 		return html.replace( rxhtmlTag, "<$1></$2>" );
 	};
 };
+
+jQuery.htmlPrefilter = function( html ) {
```
dmethvin, line 15: Restoring the legacy filter above will clobber this warning, but it's already marked `UNSAFE_` so we can do just about anything we want including format the drive.

mgol: That's true. That said, the part below with the `makeMarkup` comparison & the printing the warning could be extracted to a separate function which we could then use both here & in the `htmlPrefilter` version that `UNSAFE_restoreLegacyHtmlPrefilter` restores. What do you think?

dmethvin: I wonder if someone who uses the UNSAFE_ call actually wants the warning? Probably not.

mgol: @dmethvin I think many of them do! Some people may not care (but if they don't care about Migrate warnings they may just mute them all) but I imagine many users of this API won't be people _not_ wanting to adjust but ones that want to perform the jQuery update but who don't have time (or possibilities - they may rely on external code that needs to get updated) to deal with that breaking change at the moment. They will often still want to fix their jQuery usage & remove the `UNSAFE_` call. Migrate warnings would help them a lot here!

mgol: I think people that want this warning silenced without silencing the others & who want the old behavior restored can follow the non-Migrate based advice from our [3.5 upgrade guide](https://jquery.com/upgrade-guide/3.5/).

But I'd expect many users of this `UNSAFE_` API to want to be warned so that they can fix violations & get rid of it as it essentially undoes a security fix.

dmethvin: You're right! I could see someone restoring the filter temporarily and letting things run to see what warnings they got. Without that, their code would be broken and they might not be able to test things fully to fix them. I've added a commit, see what you think.

PR #339, opened 2020-04-14
Build: Improve ESLint configuration, check all test files as well

mgol:
The new ESLint setup is largely based on the current jQuery Core one.

### DIFF
--- a/.eslintrc-browser.json
+++ b/.eslintrc-browser.json
@@ -1,0 +1,30 @@
+{
+	"root": true,
+
+	"extends": "jquery",
+
+	// Support: IE <=9 only, Android <=4.0 only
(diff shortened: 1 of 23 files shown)


mgol:
@dmethvin Oh, I use Prettier at work and I can confirm not having to care about code formatting, just typing & pressing save to auto-format is a wonder. For jQuery, however, we've been using the existing code style for so long adding Prettier now would reformat almost every single line (Prettier code style is far different from the jQuery one) which would make blames harder. Also, one drawback of Prettier is various aspects of their imposed code style change from time to time so we'd have to reformat periodically.

Code in Migrate comes & goes so it's not that important here but being consistent with jQuery Core seems valuable to me. In the end, it's your choice, though. :)

PR #343, opened 2020-05-02
Build: Migrate sources to ES modules, use a Rollup-based build system

mgol:
The new Rollup-based build system is a very simplified version of the jQuery
one, without all the complex custom compilation aspects.

Apart from that, `grunt-compare-size` was added. While we don't care about size in Migrate to the same level as in Core, it's still good to see the impact of some changes, especially when refactoring existing code.

-1 byte compared to `master` - while the new `"use strict"` adds 1 byte, deduplication of two identical `camelCase` functions reduces the size.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,7 +2,5 @@ coverage
 external
 node_modules
 *.min.js
-src/intro.js
-src/outro.js
 test/data/jquery-*.js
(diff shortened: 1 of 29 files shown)


mgol:
I'll wait with landing this for #338 as that one is more urgent.

PR #346, opened 2020-05-03
Core: Support the jQuery slim build

### LINKED ISSUE #345: Support the jQuery slim build
While it'd be hard for Migrate to support all the possible jQuery builds,
the slim build is special as it's officially maintained by the jQuery team,
published to npm & uploaded to CDNs. It'd be worth to support it.

mgol:
While it'd be hard for Migrate to support all the possible jQuery builds,
the slim build is special as it's officially maintained by the jQuery team,
published to npm & uploaded to CDNs.

This commit adds support for the jQuery slim build in jQuery 3.0.0 & newer.
The slim build is now tested next to the regular one.

Fixes #345

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -7,6 +7,29 @@ module.exports = function( grunt ) {
 	const gzip = require( "gzip-js" );
 	const isTravis = process.env.TRAVIS;
 
+	const karmaFilesExceptJQuery = [
+		"dist/jquery-migrate.min.js",
+		"test/data/compareVersions.js",
+
+		"test/testinit.js",
+		"test/migrate.js",
+		"test/core.js",
+		"test/ajax.js",
+		"test/attributes.js",
+		"test/css.js",
+		"test/data.js",
+		"test/deferred.js",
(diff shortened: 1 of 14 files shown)


mgol:
Note that this currently doesn't pass slim builds with jQuery master because apparently the `.show()` & `.hide()` methods are erroneously excluded from it - because they were just imported in the effects module and the new Rollup-based build system follows the module dependency graph when excluding modules. I'll submit an issue to jQuery; this PR helped me found a bug there!

This issue shouldn't block us from landing this PR, though - slim tests are isolated so it's easy to see the regular ones worked. Also, jQuery 4.x is not out yet and all tests pass with both full & slim builds for all tested jQuery 3.x builds.

mgol:
@dmethvin It's hard to tell, especially that 3.x itself is still way less popular than 1.x/2.x. But I've seen enough people trying it in various places on GitHub, including it in bug reports, etc. that I think it may be more popular than we'd think! Many projects already include a Promise polyfill & they have their own AJAX library, the slim build makes a lot of sense then.

PR #350, opened 2020-05-11
Build: Update the version in package-lock.json

mgol:
So far, only the one in package.json got updated. That means
each `npm install` leaves the repository in a dirty state
by updating that field.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "jquery-migrate",
-	"version": "3.2.1-pre",
+	"version": "3.3.1-pre",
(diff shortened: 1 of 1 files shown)


timmywil:
You can also run `npm version --git-tag-version=false` to update both at once without committing/creating a tag.

dmethvin:
Yeah, sounds like I should use that rather than the manual `updatePackageVersion`. This script is pretty old but I didn't really feel like the `jquery-release` was worthwhile, particularly because it didn't work on PCs last time I tried it. (Paths were too deep.)

Issue #372, opened 2020-07-19
Detect and warn about Ajax JSON requests promoted to JSONP


dmethvin:
https://github.com/jquery/jquery/pull/4754

Since this is a security-related issue I suspect we won't want to fill it.

mgol:
What do you mean by "fill it"?

dmethvin:
Put the functionality back so it continues to work the same way it does now.

mgol:
Yeah, I don't think we want to do that here.

Now, that brings me back to the question of what's the goal of Migrate 3.x. Do we want it to work with jQuery 4 as well or will that one have Migrate 4.x? Because if that's the case then we shouldn't really polyfill any deprecated APIs in Core 3.x but only warn since they're all in source.

dmethvin:
I would like to avoid another jump like we had with Migrate 1.x where users have to use two (or three!) versions of the Migrate plugin to get to the latest jQuery. Some of the changes in jQuery 1.x made it very difficult to make a single Migrate for everything. I think Migrate 3.x should work with jQuery 4.x.

mgol:
I see, makes sense. Too bad we left some of jQuery 1.x/2.x stuff in Migrate 3.x, they'll need to stay longer.

For the sake of this issue, though, I agree, let's not bring back potential security issues, a warning should be fine. Especially that in most cases it should be easy to fix: just change `json` to `jsonp` explicitly.

mgol:
PR: #376

PR #376, opened 2020-07-22
Ajax: Warn against automatic JSON-to-JSONP promotion

mgol:
Fixes gh-372
Ref jquery/jquery#1799
Ref jquery/jquery#3376
Ref jquery/jquery#4754

### DIFF
--- a/src/jquery/ajax.js
+++ b/src/jquery/ajax.js
@@ -1,9 +1,11 @@
-import { migrateWarnFunc } from "../main.js";
+import { jQueryVersionSince } from "../compareVersions.js";
+import { migrateWarn, migrateWarnFunc } from "../main.js";
 
 // Support jQuery slim which excludes the ajax module
 if ( jQuery.ajax ) {
 
-var oldAjax = jQuery.ajax;
+var oldAjax = jQuery.ajax,
+	rjsonp = /(=)\?(?=&|$)|\?\?/;
 
 jQuery.ajax = function( ) {
 	var jQXHR = oldAjax.apply( this, arguments );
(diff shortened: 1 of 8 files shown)


test/ajax.js
```diff
@@ -19,12 +19,104 @@ QUnit.test( "jQuery.ajax() deprecations on jqXHR", function( assert ) {
...
+
+		assert.expect( 5 );
+
+		var done = assert.async(),
+			tests = [
+				function() {
+					return expectNoWarning( assert, "dataType: \"json\"",
+						function() {
+							return jQuery.ajax( {
+								url: url( "data/null.json" ),
```
mgol, line 45: This file just contains `null` which is both valid JSON & JS. Quite handy!

test/ajax.js
```diff
@@ -19,12 +19,104 @@ QUnit.test( "jQuery.ajax() deprecations on jqXHR", function( assert ) {
...
+				}
 			} );
-	} );
+		}
 
+		run( tests )
+			.then( function() {
+				done();
+			} );
+	} );
```
mgol, line 119: This is quite a lot of boilerplate but we don't have that many tests so I'm not sure if this is the moment to generalize it... If there's a way to simplify, I'm all ears. :)

test/index.html
```diff
@@ -14,7 +14,7 @@
 	<!-- Load a jQuery and jquery-migrate plugin file based on URL -->
 	<script src="testinit.js"></script>
 	<script>
-		TestManager.loadProject( "jquery", "git" );
+		TestManager.loadProject( "jquery", "3.x-git" );
```
mgol, line 17: I can separate these changes if you want.

src/jquery/ajax.js
```diff
@@ -21,4 +23,26 @@ jQuery.ajax = function( ) {
...
 
+// Only trigger the logic in jQuery <4 as the JSON-to-JSONP auto-promotion
+// behavior is gone in jQuery 4.0 and as it has security implications, we don't
+// want to restore the legacy behavior.
+if ( !jQueryVersionSince( "4.0.0" ) ) {
+
+	// Register this prefilter before the jQuery one. Otherwise, a promoted
+	// request is transformed into one with the script dataType and we can't
+	// catch it anymore.
+	jQuery.ajaxPrefilter( "+json", function( s ) {
```
mgol, line 34: The special `+` handling is not documented at https://api.jquery.com/jQuery.ajaxPrefilter/, I found it by reading source. 🙈

test/ajax.js
dmethvin: I'm fine with it, if it really gets out of hand we can refactor some other time.

test/migrate.js
```diff
@@ -1,30 +1,45 @@
...
+		// Failure; use deepEqual to show the warnings that *were* generated and the expectation
+		} else {
+			assert.deepEqual( jQuery.migrateWarnings,
+				"<warnings: " + ( expected || "1+" ) + ">", name + ": warned"
+			);
+		}
+	}
+
+	if ( result && result.then ) {
+		return jQuery.when(
```
mgol, line 35: Note to self - `jQuery.when` is unavailable in jQuery 4.x Slim. Perhaps it's better to use a Promise polyfill like [native-promise-only](https://www.npmjs.com/package/native-promise-only) that we already use in Core tests.

mgol:
@dmethvin I added one commit switching `jQuery.when` to `Promise.resolve` using a `Promise` polyfill. This works with the Core 4.x slim build. Let me know if this is still good to land.

PR #405, opened 2020-10-24
CSS: Don't throw when value is number and name is object

### LINKED ISSUE #404: jQuery.fn.css throws error when an object is used as the name, and number as the value
jQuery Migrate will treat the name as a string when a number is provided as the value. However, it shouldn't make that assumption.

jQuery UI Slider will make a call to jQuery.fn.css like so:
```
jQuery.css({left: "100%"}, 300);
```

It does this when the animate setting is specified. See https://jsfiddle.net/mweichert/784kjmoh/17/

mweichert:
Don't assume that the name is a string when the value is a number. Doing so will throw an exception.

See an example using jQuery UI Slider: https://jsfiddle.net/mweichert/784kjmoh/17/

Closes #404

### DIFF
--- a/src/jquery/css.js
+++ b/src/jquery/css.js
@@ -107,10 +107,15 @@ jQuery.fn.css = function( name, value ) {
 		} );
 	}
 	if ( typeof value === "number" ) {
-		camelName = camelCase( name );
-		if ( !isAutoPx( camelName ) && !jQuery.cssNumber[ camelName ] ) {
+		if ( typeof name === "object" ) {
 			migrateWarn( "Number-typed values are deprecated for jQuery.fn.css( \"" +
-				name + "\", value )" );
+					JSON.stringify( name ) + "\", value )" );
+		} else {
+			camelName = camelCase( name );
+			if ( !isAutoPx( camelName ) && !jQuery.cssNumber[ camelName ] ) {
+				migrateWarn( "Number-typed values are deprecated for jQuery.fn.css( \"" +
+					name + "\", value )" );
+			}
 		}
 	}
 
--- a/src/jquery/css.js
+++ b/src/jquery/css.js
@@ -105,17 +105,13 @@ jQuery.fn.css = function( name, value ) {
 		jQuery.each( name, function( n, v ) {
 			jQuery.fn.css.call( origThis, n, v );
 		} );
+		return this;
 	}
 	if ( typeof value === "number" ) {
-		if ( typeof name === "object" ) {
+		camelName = camelCase( name );
+		if ( !isAutoPx( camelName ) && !jQuery.cssNumber[ camelName ] ) {
 			migrateWarn( "Number-typed values are deprecated for jQuery.fn.css( \"" +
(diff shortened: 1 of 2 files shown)


src/jquery/css.js
```diff
@@ -107,10 +107,15 @@ jQuery.fn.css = function( name, value ) {
 		} );
 	}
 	if ( typeof value === "number" ) {
-		camelName = camelCase( name );
-		if ( !isAutoPx( camelName ) && !jQuery.cssNumber[ camelName ] ) {
+		if ( typeof name === "object" ) {
 			migrateWarn( "Number-typed values are deprecated for jQuery.fn.css( \"" +
-				name + "\", value )" );
+					JSON.stringify( name ) + "\", value )" );
```
mgol, line 112: If `name` is an object then `value` is ignored in `css` code so we shouldn't warn.

I think the bug is that we're not returning `this` in the `if` in line 104 which should already handle an object parameter. Can you try that?

You'll need to modify the test so that it mostly just asserts this doesn't crash but does its thing, just ignoring the second parameter.

test/css.js
```diff
@@ -154,3 +154,12 @@ QUnit.test( "jQuery.cssNumber", function( assert ) {
 
 	assert.ok( jQuery.cssNumber, "jQuery.cssNumber exists" );
 } );
+
+QUnit.test( "jQuery.css with object as name and number as value", function( assert ) {
```
mgol, line 158: Let's make it clear this is not public API, the second parameter should just be ignored:
```suggestion
QUnit.test( "An unsupported jQuery.fn.css(Object,Number) signature", function( assert ) {
```

mgol (changes requested):
Thanks for the PR! I added some comments.

mweichert:
@mgol I made implemented the changes suggested. I'm not sure if I followed protocol in terms of how Github reviews are conducted, as I've never used them before. I apologize if I neglected to follow procedure.

test/css.js
```diff
@@ -154,3 +154,8 @@ QUnit.test( "jQuery.cssNumber", function( assert ) {
 
 	assert.ok( jQuery.cssNumber, "jQuery.cssNumber exists" );
 } );
+
+QUnit.test( "An unsupported jQuery.fn.css(Object,Number) signature", function( assert ) {
+	assert.expect( 1 );
+	assert.ok( ( jQuery( "<div/>" ).css( { left: "100%" }, 300 ) ) );
```
mgol, line 160: Maybe make it clear we test there's no crash?
```suggestion
	assert.ok( ( jQuery( "<div/>" ).css( { left: "100%" }, 300 ), "No crash" ) );
```

mgol (approved):
Sorry for the delay. This looks good to me now, I just have one suggestion, see the comment.

@dmethvin What do you think?

mweichert:
Thanks for the help here, @mgol

mgol (approved):

dmethvin (approved):
LGTM and an easy fix, thanks!

PR #419, opened 2021-03-24
Build: Restore the external directory

mgol:
In c12e1ab660ac14cb1564d1f04121ef21e7b9094b, we removed the `external` directory
in favor of loading some files directly from `node_modules`. This works fine
locally but when deploying code for tests, this makes it impossible to not
deploy `node_modules` as well. To avoid the issue, this change restores usage
of the `external` directory.

One change is that we no longer commit this directory to the repository, its
only purpose is to have clear isolation from `node_modules`.

Ref jquery/jquery#4865

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,5 @@ CDN
 /dist/*
 !/dist/.eslintrc.json
 
-node_modules
+/external
+/node_modules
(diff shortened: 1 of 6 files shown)


mgol:
@dmethvin I merged relevant infra changes so until this lands, Migrate cannot be tested on TestSwarm.

dmethvin (approved):
Looks good. Welcome back `npmcopy`! 😄

mgol:
It looks like I missed a few things; a followup is in #420.

PR #424, opened 2021-05-01
CSS: Fix duplicate "JQMIGRATE" prefix in cssProps warning

Krinkle:
This prefix is added by `migrateWarn()`, but in the case of cssProps, the prefix was also in the string passed as parameter.


mgol (approved):
Nice catch, thanks!

Issue #436, opened 2021-06-24
$(el).data() calls break due to camelCase conversion


liady:
Currently this migration script fixes only `$.data()` calls, but `$.fn.data` still have the incompatibility with automatic camelCase conversion.
```js
// el: <div data-my-prop=test />
var data = $(el).data();
// data: {myProp: 'test'}
assert(data['my-prop']===test) // works on jQuery v2, fails on v3
```

mgol:
I have a PR adding patches for `$.fn.data`, but I've just realized it doesn't change the behavior of the shared code snippet. However, neither jQuery 1 nor jQuery 2 behave in that way either! `data-*` attribute names are camelcased in the data object.

@liady you may not remember the context after all this time, but if you do, can you explain how you got this result?

mgol:
For now, I assume #559 is the only thing needed for `data` in the 3.x line and I'm closing the issue. If we get new info, we can reopen.

PR #437, opened 2021-06-24
Add fix for $.fn.data calls

liady:
When calling `$(el).data()`, the received object contains camel-cased keys, this fixes issue #436

### DIFF
--- a/src/jquery/data.js
+++ b/src/jquery/data.js
@@ -53,7 +53,7 @@ jQuery.fn.extend( {
                         target[ prop ] === undefined
                     ) {
                         migrateWarn(
-                            "jQuery.data() always sets/gets camelCased names: " +
+                            "jQuery.fn.data() always sets/gets camelCased names: " +
                                 prop
                         );
                         return target[ camelCase( prop ) ];
@@ -64,7 +64,7 @@ jQuery.fn.extend( {
         }
         if ( arguments.length > 0 && typeof key === "string" && key !== camelCase( key ) ) {
             migrateWarn(
-                "jQuery.data() always sets/gets camelCased names: " + key
+                "jQuery.fn.data() always sets/gets camelCased names: " + key
             );
             args =
                 arguments.length > 1 ?
--- a/src/jquery/data.js
+++ b/src/jquery/data.js
@@ -45,6 +45,9 @@ jQuery.fn.extend( {
         var args, result;
         if ( arguments.length === 0 && typeof Proxy !== "undefined" ) {
             result = oldFnData.call( this );
+	    if(!result) {
+ 	      return result;
+	    }
             // eslint-disable-next-line no-undef
             return new Proxy( result, {
                 get: function( target, prop ) {
--- a/src/jquery/data.js
+++ b/src/jquery/data.js
@@ -2,6 +2,7 @@ import { migrateWarn } from "../main.js";
 import { camelCase } from "../utils.js";
 
 var oldData = jQuery.data;
+var oldFnData = jQuery.fn.data;
 
 jQuery.data = function( elem, name, value ) {
 	var curData, sameKeys, key;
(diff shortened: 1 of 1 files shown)


src/jquery/data.js
```diff
@@ -38,3 +39,39 @@ jQuery.data = function( elem, name, value ) {
 
 	return oldData.apply( this, arguments );
 };
+
+jQuery.fn.extend( {
+    data: function( key, value ) {
+        var args, result;
+        if ( arguments.length === 0 && typeof Proxy !== "undefined" ) {
+            result = oldFnData.call( this );
+            // eslint-disable-next-line no-undef
```
mgol, line 48: Why is it needed when all variables are used?

src/jquery/data.js
```diff
@@ -38,3 +39,39 @@ jQuery.data = function( elem, name, value ) {
...
+            return new Proxy( result, {
+                get: function( target, prop ) {
+                    if (
+                        prop !== camelCase( prop ) &&
+                        target[ prop ] === undefined
+                    ) {
+                        migrateWarn(
+                            "jQuery.data() always sets/gets camelCased names: " +
```
mgol, line 56: It's not `jQuery.data()` here but `jQuery.fn.data()`. The same below.

src/jquery/data.js
```diff
@@ -38,3 +39,39 @@ jQuery.data = function( elem, name, value ) {
...
+        if ( arguments.length > 0 && typeof key === "string" && key !== camelCase( key ) ) {
+            migrateWarn(
+                "jQuery.data() always sets/gets camelCased names: " + key
+            );
+            args =
+                arguments.length > 1 ?
+                    [ camelCase( key ), value ] :
+                    [ camelCase( key ) ];
+            return oldFnData.apply( this, args );
+        }
```
mgol, line 74: `$.fn.data('key-name')` already normalizes to camelCase, why is this conversion needed?

mgol (changes requested):
Thanks for the PR! I haven't tested it yet but it looks like a valid concern & a fix. I added a few comments; also:
1. Please use only tabs for indentation.
2. Please add detailed tests for this new patch.

Iqbalat87 (approved):

mgol:
@liady are you willing to finish the PR?

halimnadergrace:
Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

halimnadergrace:
Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

liady:
@mgol this is an important fix so I do want to implement the changes to this PR, however I cannot attend to it in the next 10 days.

src/jquery/data.js
liady: I got an eslint error that Proxy is not defined, could it be only for me?

src/jquery/data.js
```diff
@@ -38,3 +39,42 @@ jQuery.data = function( elem, name, value ) {
 
 	return oldData.apply( this, arguments );
 };
+
+jQuery.fn.extend( {
+    data: function( key, value ) {
```
mgol, line 44: Please use tabs for indentation, not spaces. This applies to most of the added lines.

src/jquery/data.js
```diff
@@ -1,7 +1,8 @@
 import { migrateWarn } from "../main.js";
 import { camelCase } from "../utils.js";
 
-var oldData = jQuery.data;
+var oldData = jQuery.data,
+ oldFnData = jQuery.fn.data;
```
mgol, line 5: Please use tabs for indentation, e.g. here:
```suggestion
	oldFnData = jQuery.fn.data;
```

src/jquery/data.js
```diff
@@ -38,3 +39,42 @@ jQuery.data = function( elem, name, value ) {
 
 	return oldData.apply( this, arguments );
 };
+
+jQuery.fn.extend( {
+    data: function( key, value ) {
+        var args, result;
+        if ( arguments.length === 0 && typeof Proxy !== "undefined" ) {
+            result = oldFnData.call( this );
+	    if(!result) {
```
mgol, line 48: `if` needs to be followed by the space. Please have a look at the GitHub Actions output, it shows you all the various ESLint errors of this PR.

src/jquery/data.js
mgol: Ah, this. Instead of silencing the whole rule, please add another override entry to `src/.eslintrc.json`. There's already one for `css.js`, you can copy it and just include `Proxy` as we don't use `Reflect` here.

mgol (changes requested):

mgol:
@liady Hey, could you provide a status update?

timmywil:
This PR is old and has grown stale. Please open a new one with an updated branch.

Issue #438, opened 2021-07-07
jQuery 4 warning: Number-typed values are deprecated for jQuery.fn.css ("z-index", value)


mgol:
When used against jQuery Git (future jQuery 4), there are warnings when `.css()` is used against `z-index` with a number value. This doesn't make sense as `z-index` never had `px` auto-appended. d25ecd68777e4be1185b61e07eb81dce2cfd56d6 was supposed to solve a similar problem but it only works fine for `3.x`, we need to tweak the logic.

Originally reported in https://github.com/jquery/jquery-ui/pull/1960.

mgol:
PR: #439

PR #439, opened 2021-07-07
CSS: Don't warn against `.css( "z-index", numberValue )` in jQuery >=4

mgol:
When used against jQuery Git (future jQuery 4), there were warnings when
`.css()` was used against `z-index` with a number value. This didn't make sense
as `z-index` never had `px` auto-appended. Commit
d25ecd68777e4be1185b61e07eb81dce2cfd56d6 was supposed to solve a similar problem
but it only worked fine for `3.x`.

To make the warning not show up when used with jQuery >=4 we need to fill in
`jQuery.cssNumber` with the latest version from jQuery 3.x so that the same
properties are allowed to have number values used with `.css()`.

With this PR, jQuery UI with jQuery git & jQuery Migrate shows no Migrate
warnings, similarly to what's already the case when jQuery 3.x-git is used.

Fixes gh-438

### DIFF
--- a/src/jquery/css.js
+++ b/src/jquery/css.js
@@ -81,10 +81,33 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
 	} );
 }
 
-// Create a dummy jQuery.cssNumber if missing. It won't be used by jQuery but
-// it will prevent code adding new keys to it unconditionally from crashing.
+// In jQuery >=4 where jQuery.cssNumber is missing fill it with the latest 3.x version:
+// https://github.com/jquery/jquery/blob/3.6.0/src/css.js#L212-L233
+// This way, number values for the CSS properties below won't start triggering
+// Migrate warnings when jQuery gets updated to >=4.0.0 (gh-438).
 if ( !jQuery.cssNumber ) {
-	jQuery.cssNumber = {};
+	jQuery.cssNumber = {
+		animationIterationCount: true,
+		columnCount: true,
+		fillOpacity: true,
+		flexGrow: true,
+		flexShrink: true,
+		fontWeight: true,
+		gridArea: true,
+		gridColumn: true,
+		gridColumnEnd: true,
+		gridColumnStart: true,
+		gridRow: true,
+		gridRowEnd: true,
+		gridRowStart: true,
+		lineHeight: true,
+		opacity: true,
+		order: true,
+		orphans: true,
+		widows: true,
+		zIndex: true,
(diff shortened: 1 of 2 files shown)


mgol:
With this PR, jQuery UI with jQuery git & jQuery Migrate shows no Migrate warnings, similarly to what's already the case when jQuery 3.x-git is used.

test/css.js
```diff
@@ -62,7 +62,7 @@ QUnit.test( "jQuery.css with arrays", function( assert ) {
 
 QUnit.test( "jQuery.css with numbers", function( assert ) {
 	var jQuery3OrOlder = compareVersions( jQuery.fn.jquery, "4.0.0" ) < 0,
-		whitelist = [
+		allowlist = [
```
dmethvin, line 65: 👍

src/jquery/css.js
```diff
@@ -81,10 +81,33 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
...
 
-// Create a dummy jQuery.cssNumber if missing. It won't be used by jQuery but
-// it will prevent code adding new keys to it unconditionally from crashing.
+// In jQuery >=4 where jQuery.cssNumber is missing fill it with the latest 3.x version:
+// https://github.com/jquery/jquery/blob/3.6.0/src/css.js#L212-L233
+// This way, number values for the CSS properties below won't start triggering
+// Migrate warnings when jQuery gets updated to >=4.0.0 (gh-438).
 if ( !jQuery.cssNumber ) {
-	jQuery.cssNumber = {};
+	jQuery.cssNumber = {
```
dmethvin, line 89: Man we can never win. 😸

dmethvin (approved):

src/jquery/css.js
mgol: This is not 100% necessary as Core doesn't need this object in 4.x; we could as well just enumerate allowlisted properties in Migrate. However, by doing it via `jQuery.cssNumber` we have the same Migrate code path for all jQuery versions and it's easier to copy the object from the `3.x-stable` Core branch if there are any changes.

dmethvin: Yeah it seems easier to just copy over the list. It's an exposed object so who knows, maybe some software out there is looking at it and not just setting properties.

mgol:
Landed. I don't think we need to release at the moment as it only affects jQuery Git at the moment; by the time we release jQuery 4.0 we're likely to have another Migrate release for different reasons anyway.

PR #445, opened 2021-12-01
Build: Switch from Travis to GitHub Actions

mgol:


mgol:
I don't know why the actions don't run but we had the same issue in jQuery UI and I was only able to unlock it when the PR was merged.

mgol:
Let me try merging this & submitting a PR, we'll see if it works...

mgol:
The `main` build run successfully: https://github.com/jquery/jquery-migrate/actions/runs/1531830986

PR #459, opened 2022-03-24
Build: Make the release script an ES module

mgol:
This is now required by the latest chalk versions.


mgol:
The script works, I just used it. :)

PR #470, opened 2022-08-08
Core: Accept only jQuery 3.x-4.x

mgol:


dmethvin (approved):
The table definitely helps!

PR #479, opened 2023-01-15
Build: Downgrade uglify-js to fix tests in IE 9

mgol:
We can upgrade for Migrate 4.0. It's not worth spending time on figuring out how to fix IE 9 with new UglifyJS at this point.

The `ie` option had to be renamed to `ie8` as the older version doesn't support the former. In Migrate 4 we can remove it altogether.


mgol:
This shouldn't be controversial so I'm going to merge this without waiting for a review.

PR #482, opened 2023-01-30
Docs: Mark `jQuery.fn.load( [ eventData ], handler )` as removed

jp41011:
I think that jquery.fn.unload() should have the "is deprecated and removed" label. .unload() was deprecated in jquery v1.8 and removed in v3.0 So if you get this "warning" it will not actually work once you remove the jquery migrate plugin.

Reference: https://api.jquery.com/unload/

### DIFF
--- a/warnings.md
+++ b/warnings.md
@@ -61,7 +61,7 @@ This is _not_ a warning, but a console log message the plugin shows when it firs
 **Solution:** Change any use of `$().error(fn)` to `$().on("error", fn)`.
 
 ### \[shorthand-removed-v3\] JQMIGRATE: jQuery.fn.load() is deprecated
-### \[shorthand-removed-v3\] JQMIGRATE: jQuery.fn.unload() is deprecated
+### \[shorthand-removed-v3\] JQMIGRATE: jQuery.fn.unload() is deprecated and removed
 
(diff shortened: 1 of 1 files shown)


mgol:
Thanks. Please sign our CLA.

warnings.md
```diff
@@ -61,7 +61,7 @@ This is _not_ a warning, but a console log message the plugin shows when it firs
 **Solution:** Change any use of `$().error(fn)` to `$().on("error", fn)`.
 
 ### \[shorthand-removed-v3\] JQMIGRATE: jQuery.fn.load() is deprecated
-### \[shorthand-removed-v3\] JQMIGRATE: jQuery.fn.unload() is deprecated
+### \[shorthand-removed-v3\] JQMIGRATE: jQuery.fn.unload() is deprecated and removed
```
mgol, line 64: `load` should be marked as removed as well - as this particular signature did get removed: https://api.jquery.com/load-event/. What was left is https://api.jquery.com/load/; this is a different API that just shared the same name. Which one was to be used was detected by checking arguments passed to the method.

I just wonder how to do this to not confuse people about the non-event `load` method being removed... Perhaps by changing the name above to `jQuery.fn.load( [eventData ], handler )`?

mgol (changes requested):
Thanks for signing the CLA. I missed it as signing the CLA doesn't send GitHub notifications in itself.

I added one comment.

warnings.md
jp41011: I added a commit with the suggested change to the `load()` reference. 
I agree that does make it easier to distinguish between the two loads.
Also we do mention the AJAX-related `.load()` in the **Cause:** section, so that mention along with the change to the method call hopefully is enough to highlight the difference between the two loads. 
What do you think?

mgol: Yes, that sounds good to me. Thanks!

jp41011: Cool. Thanks.

Issue #471, opened 2022-08-11
Only warn against `removeAttr(booleanAttr)` if property value not `false`


mgol:
Migrate warns against `.removeAttr()` called on boolean attributes as this used to set the associated property to `false` but it doesn't do so anymore since jQuery 3.0.0.

However, this is only an issue if the property wasn't already set to `false`. There are valid reasons to call `removeAttr` on boolean attributes and it shouldn't cause issues as long as the property is being updated first.

This is not just a theoretical concern - AngularJS sets both boolean properties & then associated attributes, calling `removeAttr` if the value passed was `false`. See a report about that at https://github.com/angular/angular.js/issues/16877. I also hit it at my employer when updating jQuery.

The proposal is to read the property and if it's already `false`, skip the warning.

PR #484, opened 2023-02-22
Attributes: Don't warn against `removeAttr` if property `false`

mgol:
Some frameworks, like AngularJS, invoke `removeAttr` on boolean attributes but only after setting the respective property to `false`. Such usage is not dangerous and would continue to work even after Migrate is removed. Therefore, only warn for `removeAttr(booleanAttribute)` if the property wasn't `false` already.

Fixes gh-471

### DIFF
--- a/src/jquery/attributes.js
+++ b/src/jquery/attributes.js
@@ -5,10 +5,24 @@ var oldRemoveAttr = jQuery.fn.removeAttr,
 	rmatchNonSpace = /\S+/g;
 
 migratePatchFunc( jQuery.fn, "removeAttr", function( name ) {
-	var self = this;
+	var self = this,
+		patchNeeded = false;
 
 	jQuery.each( name.match( rmatchNonSpace ), function( _i, attr ) {
 		if ( jQuery.expr.match.bool.test( attr ) ) {
+
+			// Only warn if at least a single node had the property set to
+			// something else than `false`. Otherwise, this Migrate patch
+			// doesn't influence the behavior and there's no need to set or warn.
+			self.each( function() {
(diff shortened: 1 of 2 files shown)


src/jquery/attributes.js
```diff
@@ -5,10 +5,24 @@ var oldRemoveAttr = jQuery.fn.removeAttr,
...
+		patchNeeded = false;
 
 	jQuery.each( name.match( rmatchNonSpace ), function( _i, attr ) {
 		if ( jQuery.expr.match.bool.test( attr ) ) {
+
+			// Only warn if at least a single node had the property set to
+			// something else than `false`. Otherwise, this Migrate patch
+			// doesn't influence the behavior and there's no need to set or warn.
+			self.each( function() {
+				if ( jQuery( this ).prop( attr ) !== false ) {
```
dmethvin, line 18: Could we just use `this.attr !== false` here?

dmethvin (approved):
Sounds reasonable to skip this if the attr is already false!

src/jquery/attributes.js
mgol: I was playing safe to not omit any possible prop hooks. We _almost_ don't have any for boolean props... except for `selected` in IE:
https://github.com/jquery/jquery/blob/3.6.3/src/attributes/prop.js#L91-L125
It seems we should use the jQuery API then, shouldn't we?

dmethvin: Yeah I suppose we might add a prop hook so this is the safest way. 👍

dmethvin (approved):

Issue #495, opened 2024-01-30
jquery-migrate: jQuery 4.0 error: `jQuery.expr.match.bool` does not exist


nagyimre1980:
jquery-migrate has been incompatible with jquery version 4.0 for a few weeks.
"Query.expr.match.bool" does not exist.
The following part was removed from jquery 4.0 a few weeks ago. "bool: new RegExp( "^(?:" + booleans + ")$", "i" ),"
(jquery 4.0: https://releases.jquery.com/git/jquery-git.js)

migratePatchFunc( jQuery.fn, "removeAttr", function( name ) {
	var self = this,
		patchNeeded = false;

	jQuery.each( name.match( rmatchNonSpace ), function( _i, attr ) {
		if ( jQuery.expr.match.bool.test( attr ) ) {

mgol:
I am aware of this issue, we're seeing it in our tests at https://swarm.jquery.org/project/jquerymigrate. There's no urgency since Git versions of jQuery & compatibility with them is not supported but we'll fix it.

I haven't created an issue for it so we can use this one for tracking.

mgol:
For the record, the cause is https://github.com/jquery/jquery/pull/5384

mgol:
Fix at #496.

mgol:
Released as part of jQuery Migrate `3.5.0`.

mgol:
Released as part of jQuery Migrate `3.5.0`: https://github.com/jquery/jquery-migrate/releases/tag/3.5.0.

PR #496, opened 2024-02-03
Attributes: Fix compatibility with jQuery 4.x

mgol:
In jquery/jquery#5384, `jQuery.expr.match.bool` stopped being defined and the regex matching boolean attributes is no longer exposed. Inline it in Migrate to avoid it crashing on jQuery 4.x.

Fixes gh-495


dmethvin (approved):
Yep, easiest way to fix this!

PR #503, opened 2024-03-18
Replace karma and testswarm with jQuery test runner; add filestash workflow

timmywil:
- replaces karma and testswarm with the new jQuery test runner
- add GH workflow for updating git files on filestash
- selenium is having an issue starting safaridriver when loading esmodules. Retries succeed, but inconsistently. Leaving that out for now.
- intentionally running both dev and min on every supported jQuery version. Before, swarm was only running min on certain versions, but this was actually easier to support and didn't add much time to the test runs.
- environment variables are already in place
- upgrade all dependencies, but lock uglify to a version that supports the `ie8` argument.

**Disable jenkins job before merging**

Sample runs: 

Filestash dry run: https://github.com/timmywil/jquery-migrate/actions/runs/8328770535/job/22789518882
Browserstack: https://github.com/timmywil/jquery-migrate/actions/runs/8328770533
Browserstack (git): https://github.com/timmywil/jquery-migrate/actions/runs/8328770540


.github/workflows/browserstack.yml
```diff
@@ -0,0 +1,87 @@
+name: Browserstack
```
mgol, line 1: Let's name this `BrowserStack Core 3.x` or something similar and let's rename the file as well. We will need a separate file for Core 4.x as soon as one is released and this will result in less renaming as we'll support both for at least one release with different browser sets. And even if we align browser sets when Migrate 4.0 gets released, it still makes sense to separate Core major lines to avoid one big run.

.github/workflows/browserstack.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Build
+        run: npm run build
+
+      - name: Test
+        run: |
+          npm run test:unit -- \
+            -v --browserstack "${{ matrix.BROWSER }}" \
+            --jquery-migrate dev --jquery-migrate min \
+            --jquery 3.x-git --jquery 3.x-git.min --jquery 3.x-git.slim --jquery 3.x-git.slim.min \
+            --jquery 3.7.0 --jquery 3.7.0.slim --jquery 3.6.4 --jquery 3.6.4.slim \
```
mgol, line 82: We should use `3.7.1` now:
```suggestion
            --jquery 3.7.1 --jquery 3.7.1.slim --jquery 3.6.4 --jquery 3.6.4.slim \
```
but perhaps it's worth doing it in a separate PR as it needs to be changed in multiple places.

.github/workflows/node.js.yml
```diff
@@ -1,37 +1,75 @@
...
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
+    name: ${{ matrix.MIGRATE_VERSION }} - jQuery ${{ matrix.JQUERY_VERSION }} (Chrome/Firefox)
+    strategy:
+      fail-fast: false
+      matrix:
+        MIGRATE_VERSION: ["min"]
+        JQUERY_VERSION: ["3.x-git.min", "3.x-git.slim.min"]
```
mgol, line 19: Can we test against `git.min` & `git.slim.min` as well? It shouldn't be too long and it'd be great to catch potential issues directly at the PR level.

If you change it here then remember also about the Safari entry.

Gruntfile.js
```diff
@@ -69,42 +28,13 @@ module.exports = function( grunt ) {
 				cache: "build/.sizecache.json"
 			}
 		},
-		tests: {
```
mgol, line 72: Can we migrate other tasks as well? It can happen in a separate PR. Migrate should be on a similar maintenance level as Core so it'd be good to get rid of Grunt here as well.

package.json
```diff
@@ -1,88 +1,93 @@
...
-			"Manipulation",
-			"Offset",
-			"Queue",
-			"Selector",
-			"Serialize",
-			"Traversing",
-			"Wrap"
-		]
-	}
+  "name": "jquery-migrate",
```
mgol, line 2: I know it aligns us with the current Core but is this a good change? npm is able to detect indentation of `package.json` and using tabs is more consistent with the jQuery code style.

package.json
```diff
@@ -1,88 +1,93 @@
...
+  "bugs": {
+    "url": "http://bugs.jquery.com/"
+  },
+  "license": "MIT",
+  "scripts": {
+    "build": "grunt",
+    "test:browser": "npm run build && npm run test:unit -- -b chrome -b firefox -h",
+    "test:safari": "npm run build && npm run test:unit -- -v -b safari",
+    "test:server": "node test/runner/server.js",
+    "test:unit": "node test/runner/command.js"
```
mgol, line 25: Can you also add a `test` script that does everything that's sensible to do locally? It's customary to expect `npm install && npm test` - or its short form `npm it` to install dependencies, run the build & all tests that should verify your change didn't break anything. We also have this in Core.

`npm test` should not run Safari tests so that it can be run on Windows.

package.json
```diff
@@ -1,88 +1,93 @@
...
+    "grunt-contrib-uglify": "4.0.1",
+    "grunt-contrib-watch": "1.1.0",
+    "grunt-eslint": "24.0.0",
+    "grunt-npmcopy": "0.2.0",
+    "gzip-js": "0.3.2",
+    "load-grunt-tasks": "5.1.0",
+    "native-promise-only": "0.8.1",
+    "qunit": "2.19.1",
+    "rollup": "2.70.1",
+    "selenium-webdriver": "4.18.1",
```
mgol, line 57: Can we update Rollup as well? Or is IE 9 support an issue?

test/runner/.eslintrc.json
```diff
@@ -0,0 +1,34 @@
...
+
+	"overrides": [
+		{
+			"files": ["**/*"],
+      "globals": {
+        "fetch": false
+      },
+			"parserOptions": {
+        "ecmaVersion": 2022,
+				"sourceType": "module"
```
mgol, line 14: Indentation is off in this file.

test/runner/.eslintrc.json
```diff
@@ -0,0 +1,34 @@
+{
+	"root": true,
```
mgol, line 2: Can we migrate ESLint to the flat config in Migrate as well?

test/runner/command.js
```diff
@@ -0,0 +1,145 @@
...
+		default: [ "chrome" ]
+	} )
+	.option( "headless", {
+		alias: "h",
+		type: "boolean",
+		description:
+			"Run tests in headless mode. Cannot be used with --debug or --browserstack.",
+		conflicts: [ "debug", "browserstack" ]
+	} )
+	.option( "concurrency", {
```
mgol, line 56: The `esm` option should be supported here as well, just like in Core.

test/runner/command.js
```diff
@@ -0,0 +1,145 @@
...
+	.help().argv;
+
+if ( typeof argv.listBrowsers === "string" ) {
+	listBrowsers( buildBrowserFromString( argv.listBrowsers ) );
+} else if ( argv.stopWorkers ) {
+	stopWorkers();
+} else if ( argv.browserstackPlan ) {
+	console.log( await getPlan() );
+} else {
+	run( argv );
```
mgol, line 144: In Core, you had:
```js
	run( {
		...argv,
		browsers: argv.browser,
		modules: argv.module
	} );
```
Is that not needed here? If so, why?

test/runner/createTestServer.js
```diff
@@ -0,0 +1,65 @@
...
+
+	// Serve static files
+	app.use( "/dist", express.static( "dist" ) );
+	app.use( "/src", express.static( "src" ) );
+	app.use( "/test", express.static( "test" ) );
+	app.use( "/external", express.static( "external" ) );
+
+	// Serve development jQuery, which is expected to be
+	// a sibling of the jquery-migrate repo.
+	app.use( "/jquery", express.static( "../jquery" ) );
```
mgol, line 62: I have my main jQuery repo in the `jquery-core` directory so that I can differentiate it from its parent directory containing all the jQuery stuff, which is called... `jquery`.

I guess I can rename that parent directory in addition to the main jQuery one... It's going to be a bit painful but maintaining the upstream repo name does seem to be the most natural choice.

test/runner/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
mgol, line 2: Can we put that at the top level as in Core? `Gruntfile.js` can be renamed to `Gruntfile.cjs` for now.

test/runner/lib/buildTestUrl.js
```diff
@@ -0,0 +1,32 @@
...
+	if ( !port ) {
+		throw new Error( "No port specified." );
+	}
+
+	const query = new URLSearchParams();
+	for ( const module of modules ) {
+		query.append( "moduleId", generateModuleId( module ) );
+	}
+
+	if ( jquery ) {
```
mgol, line 16: Let's support the `esm / esmodules` option pair here as in Core.

mgol (commented):
A few comments. Can't wait to see it in Migrate!

.github/workflows/browserstack.yml
timmywil: I think we can go ahead and make that change.

Gruntfile.js
timmywil: Yes, but that might come a good bit later. My first priority is getting jenkins decommissioned.

package.json
timmywil: I think we've sort of always had an implicit preference towards spaces in JSON. It comes out in different places. For instance, when we write JSON, we often do `JSON.stringify( contents, null, '  ')`, perhaps because that's what comes to us naturally. jQuery code style doesn't necessarily have to apply to the JSON.

package.json
timmywil: I think it will just be an alias for `test:browser` right now, until we migrate all grunt tasks.

test/runner/.eslintrc.json
timmywil: yes, eventually

test/runner/command.js
timmywil: I don't think it should. Core sets the version of one library. It uses top level options and parameters to set the core version. Conversely, migrate always sets the versions of two distinct libraries. Instead, esm has been supported in migrate by passing `esmodules` as a *value* to the `plugin` URL parameter, and is supported here as a value of the `jquery-migrate` option.

test/runner/command.js
timmywil: With the extra options (`jquery` and `jqueryMigrate`) also getting pluralized, I moved the pluralization to the run parameters instead so the command can be agnostic. I could do the same in core, but it became more necessary with the extra arguments.

test/runner/.eslintrc.json
timmywil: Ok, this is a good argument for not changing all the JSON indentation, at least not yet. Maybe when we migrate off grunt and use eslint flat config, we could introduce spaces in JSON if we agree.

package.json
mgol: I'm fine with that. Still, it's good to have the `test` script defined.

I think you could also add linting to that, though.

test/runner/.eslintrc.json
mgol: OK, I assume you mean a separate PR some time later.

test/runner/.eslintrc.json
mgol: I think I'd prefer tabs in JSON as well for internal code style consistency. Not a strong preference but it feels to me like a better choice here.

test/runner/createTestServer.js
timmywil: We could make it an optional environment variable

test/runner/.eslintrc.json
mgol: But yeah, whatever we settle on can be done separately. I mostly care about here that it's at least consistent within a single file. 😉

test/runner/command.js
mgol: Good point. No change needed then.

test/runner/lib/buildTestUrl.js
timmywil: See https://github.com/jquery/jquery-migrate/pull/503#discussion_r1530491173

package.json
timmywil: `test:browser` includes `npm run build`, which runs `grunt`, which includes `grunt lint`.

package.json
timmywil: I was thinking of sinon. Rollup might be fine at least.

test/runner/createTestServer.js
timmywil: I went ahead and added the env var just in case.

src/jquery/css.js
```diff
@@ -81,7 +81,7 @@ if ( jQueryVersionSince( "3.4.0" ) && typeof Proxy !== "undefined" ) {
 }
 
 // In jQuery >=4 where jQuery.cssNumber is missing fill it with the latest 3.x version:
-// https://github.com/jquery/jquery/blob/3.7.0/src/css.js#L216-L246
+// https://github.com/jquery/jquery/blob/3.7.1/src/css.js#L216-L246
```
timmywil, line 84: I did check the link here and the lines were the same.

timmywil:
New sample runs:

3.x: https://github.com/timmywil/jquery-migrate/actions/runs/8345191538
git: https://github.com/timmywil/jquery-migrate/actions/runs/8345191528

test/runner/command.js
mgol: I see, makes sense. Maybe you could also do it in Core later for consistency.

mgol (approved):
Looks great, thanks!

mgol:
@timmywil I disabled both Jenkins Migrate jobs. Feel free to merge the PR.

PR #508, opened 2024-04-04
Tests: share queue/browser handling for all worker types 

timmywil:
- one queue to rule them all: browserstack and selenium
- retries and hard retries are now supported in selenium
- selenium tests now re-use browsers in the same way as browserstack
- also fix concurrency so no jobs are cancelled

Companion PR for https://github.com/jquery/jquery/pull/5460 and https://github.com/jquery/jquery/pull/5465

#### Sample test runs

3.x: https://github.com/timmywil/jquery-migrate/actions/runs/8548231361
git: https://github.com/timmywil/jquery-migrate/actions/runs/8548231364


mgol:
For the record, I reviewed this by doing a directory diff on the jQuery version of `test/runner` and the one from this PR using [Kaleidoscope](https://kaleidoscope.app/).

timmywil:
I'm going to keep the separate commits by rebasing, but I've added a reference to each commit message to this PR.

PR #512, opened 2024-05-30
Build: replace grunt tasks with npm scripts

timmywil:
Includes:

- **lint**: lint all code using new eslint flat config
- **build**: build using rollup, minify using uglify, process for dist the same as core
- **npmcopy**: copy some node module files to the external folder
- **start**: watch source files and rebuild on changes using rollup.watch

Also:
- Update the build function in the release script to use the new build
- Update CONTRIBUTING.md and README.md with new scripts
- Confirm the min file was identical to the main branch build, aside from the version update.
- Keep uglify-js at 3.9.4, which is the last version that officially supported the `ie8` argument, which ironically we need for ie9
- add schedules to browserstack workflows so they run once a week
- update browserstack project name in the test runner to jquery-migrate (it had been overridden when copying test queue code over from jquery)

Sample BrowsersStack runs:

git: https://github.com/timmywil/jquery-migrate/actions/runs/9304840969
3.x: https://github.com/timmywil/jquery-migrate/actions/runs/9305649366


eslint.config.js
```diff
@@ -0,0 +1,195 @@
...
+			"no-unused-vars": [
+				"error",
+				{ args: "after-used", argsIgnorePattern: "^_" }
+			]
+		}
+	},
+
+	{
+		files: [ "test/data/**" ],
+		ignores: [ "test/data/jquery-2.2.3.js", "test/data/qunit-start.js" ],
```
mgol, line 118: Can we make test jQuery updates easier by changing this to:
```suggestion
		ignores: [ "test/data/jquery-*.js", "test/data/qunit-start.js" ],
```
?

eslint.config.js
```diff
@@ -0,0 +1,195 @@
...
+		rules: {
+			...jqueryConfig.rules,
+			strict: [ "error", "function" ]
+		}
+	},
+
+	{
+		files: [ "dist/jquery-migrate.js" ],
+		languageOptions: {
+			ecmaVersion: 5,
```
mgol, line 175: We also need to verify the minified file is ES5.

package.json
```diff
@@ -42,24 +46,14 @@
...
-		"selenium-webdriver": "4.18.1",
+		"qunit": "2.21.0",
+		"rollup": "4.18.0",
+		"selenium-webdriver": "4.21.0",
+		"uglify-js": "3.9.4",
 		"yargs": "17.7.2"
 	},
-	"overrides": {
-		"uglify-js": "3.9.4"
-	},
```
mgol, line 62: It's great to get rid of this.

src/jquery/core.js
```diff
@@ -130,18 +130,18 @@ if ( jQueryVersionSince( "3.3.0" ) ) {
...
+		return ( type === "number" || type === "string" ) &&
 
 				// parseFloat NaNs numeric-cast false positives ("")
 				// ...but misinterprets leading-number strings, e.g. hex literals ("0x...")
 				// subtraction forces infinities to NaN
 				!isNaN( obj - parseFloat( obj ) );
-		}, "isNumeric",
-		"jQuery.isNumeric() is deprecated"
+	}, "isNumeric",
+	"jQuery.isNumeric() is deprecated"
```
mgol, line 144: This looks weird with the next line - the closing paren is at the same level as its last entry. Should we merge a few lines here or do something else to prevent this?

test/unit/jquery/effects.js
```diff
@@ -10,7 +10,7 @@ QUnit.test( "jQuery.easing", function( assert ) {
 
 	assert.expect( 4 );
 
-	jQuery.easing.testOld = function( p, n, firstNum, diff ) {
+	jQuery.easing.testOld = function( _p, _n, _firstNum, _diff ) {
```
mgol, line 13: Perhaps we should just delete the arguments since they're not used.

mgol (changes requested):
A few comments.

build/tasks/minify.js
```diff
@@ -0,0 +1,81 @@
...
+				// See https://github.com/mishoo/UglifyJS2/issues/2198
+				typeofs: false
+			},
+			output: {
+				ascii_only: true,
+
+				// Support: Android 4.0 only
+				// UglifyJS 3 breaks Android 4.0 if this option is not enabled.
+				// This is in lieu of setting ie for all of mangle, compress, and output
+				ie8: true,
```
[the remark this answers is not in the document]
timmywil: That sounds fine, but I say we switch the minify task off of uglify later and keep changes to a minimum until after the 4.0 release. All we're doing here is removing grunt.

eslint.config.js
```diff
@@ -0,0 +1,195 @@
...
+					]
+				}
+			]
+		}
+	},
+
+	{
+		files: [ "test/unit/**" ],
+		languageOptions: {
+			ecmaVersion: 2021,
```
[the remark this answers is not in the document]
timmywil: I don't remember. 2015 is fine.

eslint.config.js
timmywil: Do we? We don't usually lint minified files and the old config was ignoring them. I think we can safely say if the dev version is es5, the minified one is too. And any odd variance would be caught by the browserstack runs.

src/jquery/core.js
timmywil: I just ran prettier over it. Let me know what you think.

test/unit/jquery/effects.js
timmywil: I tried that, but the number of arguments matters for the test.

timmywil: I'll add a comment.

eslint.config.js
timmywil: I'll go ahead and do this since we're going to do this in jQuery core.

build/tasks/minify.js
```diff
@@ -0,0 +1,80 @@
...
+
+				// Support: IE <11
+				// typeofs transformation is unsafe for IE9-10
+				// See https://github.com/mishoo/UglifyJS2/issues/2198
+				typeofs: false
+			},
+			output: {
+				ascii_only: true,
+
+				// Support: Android 4.0 only, IE9 only
```
mgol, line 47:
```suggestion
				// Support: Android 4.0 only, IE 9 only
```

src/jquery/core.js
```diff
@@ -11,68 +11,87 @@ var findProp,
...
-
-		// JQuery( "#" ) is a bogus ID selector, but it returned an empty set
-		// before jQuery 3.0
-		migrateWarn( "selector-empty-id", "jQuery( '#' ) is not a valid selector" );
-		args[ 0 ] = [];
-	}
+migratePatchFunc(
+	jQuery.fn,
+	"init",
+	function( arg1 ) {
```
mgol, line 24: I'd like to avoid all this reformatting in this file. We don't use Prettier and any time we do major code style changes like that - especially ones modifying line numbers - it gets harder to look at history.

If something is a result of stricter ESLint then fine, but this looks excessive to me.

mgol (changes requested):

src/jquery/core.js
timmywil: I don't really know of a better way to format this. I understand the concern about diffing, but I think the way Prettier formatted this is the way it should have been formatted from the beginning. And, doing it now makes it less likely we'll have to make more formatting changes in the future. It was ESLint that prompted this, but I prefer to let tools do the formatting as much as possible so that maintainers like us don't waste any time doing delicate dances to get the code to pass lint.

mgol: Well, there is also the current style. Right now, this PR splits the style as other files use what was here before; you'd have to reformat everything to make it consistent again.

What exactly in ESLint triggered these changes?

> I prefer to let tools do the formatting as much as possible so that maintainers like us don't waste any time doing delicate dances to get the code to pass lint

I have a more nuanced view on this. I do use formatters a lot - and usually it's Prettier as well - as it helps with avoiding some discussions, especially in large teams. But integrating such a tool doesn't mean you stop having formatting changes, unfortunately. Prettier's GitHub has many discussions about how to best format various code constructs and they change things from time to time. In my experience, having Prettier in a project means that you periodically need to update it & then reformat a huge percentage of files, creating huge diffs and obstructing the Git history. In larger teams & projects I believe it's still worth it but I have doubts about jQuery with just a few regular committers and quite rare source changes. In some files, we'd end up with the only modifications being occasional Prettier runs.

One option is to employ usage of https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view, although I expect it's not perfect and won't erase all the noise.

As for the current Migrate style as it's on `main`, I actually like it as the following:
```js
migratePatchFunc( jQuery.fn, "init", function( arg1 ) {
	/* code */
}, "selector-empty-id" );
```
is quite similar visually to:
```js
jQuery.fn.init = function( arg1 ) {
	/* code */
};
```
which is mostly what happens, modulo the `migratePatchFunc` wrapper that is needed for being able to selectively disable patches.

timmywil: The first is what was failing. FWIW, I disagree on which is better. In the first example, it's easy to overlook that last argument after the function, especially when the function has enough lines. Putting each argument on its own line is the clearest way to prevent that. And, there are several cases in this file where there are 2 string arguments after the function, exacerbating the problem.

But, if you're adamant, we can avoid changing the whole file. This all started because you preferred to not change the indentation in the one of the functions that was inconsistent with the others, but had a ternary return that made it look awkward to be consistent. A possible solution for that is to do what was done elsewhere and start the function declaration on the next line. See the latest.

mgol (approved):
Thanks, that looks less noisy, LGTM.

As for the code style itself, I could be convinced what's better in either direction but I tend to skew towards trying to avoid changing the existing style too much, at least as long as we don't employ automatic tooling that totally enforces that.

But that would have its own drawbacks in the context of jQuery projects as I mentioned in one of my comments. And I don't currently have a good answer for that.

Issue #474, opened 2022-08-18
Warn against jQuery.UNSAFE_restoreLegacyHtmlPrefilter usage


mgol:
This is unusual as it's not the Core API but I think we should warn against usage of `jQuery.UNSAFE_restoreLegacyHtmlPrefilter` and recommend calling `jQuery.migrateEnablePatches( "self-closed-tags" )` instead. I'd like to drop this method in Migrate 4.x.

PR #518, opened 2024-07-08
Manipulation: Deprecate jQuery.UNSAFE_restoreLegacyHtmlPrefilter

mgol:
This is unusual as it's not the Core API but we support this patch via a more generic API now:

```
jQuery.migrateEnablePatches( "self-closed-tags" )
```

Fixes gh-474


timmywil (approved):

mgol:
A docs follow-up in #526.

PR #531, opened 2024-09-16
Ajax: Fill in & warn against automatic JSON-to-JSONP promotion

mgol:
So far, the patch was only warning about the automatic promotion, but it wasn't filling the behavior back to jQuery 4+. This has been fixed.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -130,6 +130,7 @@ export default [
 				QUnit: false,
 				url: true,
 				compareVersions: true,
+				jQueryVersionSince: false,
 				expectWarning: true,
 				expectNoWarning: true,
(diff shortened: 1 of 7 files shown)


eslint.config.js
```diff
@@ -130,6 +130,7 @@ export default [
 				QUnit: false,
 				url: true,
 				compareVersions: true,
+				jQueryVersionSince: false,
```
timmywil, line 133: Why wasn't this an issue before?

timmywil (approved):

eslint.config.js
mgol: This API was not used in `test/data/*` files previously.

PR #533, opened 2024-09-17
Build: Test on iOS 18

mgol:
1. Test on iOS 18 with jQuery 3.x & Git.
2. Don't test on iOS 15 with jQuery Git.


timmywil:
I think we should remove iOS 15 from 3.x as well.

timmywil (approved):

Issue #504, opened 2024-03-18
Restore boolean attribute treatment from 3.x


mgol:
See https://github.com/jquery/api.jquery.com/issues/1243 & https://github.com/jquery/jquery/pull/5452 for changes planned for 4.0. We should restore special boolean attributes treatment and warn if the setter is used with an unrecognized value. In the getter, warn if the value is different than the lowercased attribute name.

Remember to go through the list of boolean attributes in the spec and exempt any new values (like `until-found` for the `hidden` attribute) from such a conversion. New attributes should not be added to the list but any changes in existing ones will need to be accounted for.

mgol:
I have this work in progress but I'm pausing it for now to focus on other priorities; I'll return to this before the jQuery 4.0.0 RC.

mgol:
PR: #540

PR #540, opened 2024-10-08
Attributes: Restore boolean attribute & false setter treatment from 3.x

mgol:
Restore & warn against:
* boolean attributes set to something different than their lowercase names
* boolean attributes queried when set to something different than their lowercase names
* non-boolean non-ARIA attributes set to `false`

Fixes gh-504
Ref jquery/jquery#5452
Ref jquery/api.jquery.com#1243

### DIFF
--- a/src/jquery/attributes.js
+++ b/src/jquery/attributes.js
@@ -1,9 +1,116 @@
 import { migratePatchFunc, migrateWarn } from "../main.js";
+import { jQueryVersionSince } from "../compareVersions.js";
 
 var oldRemoveAttr = jQuery.fn.removeAttr,
+	oldJQueryAttr = jQuery.attr,
 	oldToggleClass = jQuery.fn.toggleClass,
-	rbooleans = /^(?:checked|selected|async|autofocus|autoplay|controls|defer|disabled|hidden|ismap|loop|multiple|open|readonly|required|scoped)$/i,
-	rmatchNonSpace = /\S+/g;
+	booleans = "checked|selected|async|autofocus|autoplay|controls|defer|" +
+		"disabled|hidden|ismap|loop|multiple|open|readonly|required|scoped",
+	rbooleans = new RegExp( "^(?:" + booleans + ")$", "i" ),
+	rmatchNonSpace = /\S+/g,
+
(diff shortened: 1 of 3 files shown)


src/jquery/attributes.js
```diff
@@ -1,9 +1,116 @@
...
+							"' value is different from its lowercased name" );
+
+					// jQuery <4 attr hooks setup is complex: there are attr
+					// hooks, bool hooks and selector attr handles. Only
+					// implement the logic in jQuery >=4 where it's missing
+					// and there are only attr hooks.
+					if ( jQueryVersionSince( "4.0.0" ) ) {
+						return name.toLowerCase();
+					}
+					return null;
```
mgol, line 46: This works in jQuery 3.x because - while `boolHooks` being a fallback for boolean attributes if `attrHooks[ name ]` didn't exist would not get picked up here - `boolHooks` only handle a setter and we're in a getter here. The getter - if the `get` attr hook is missing or returns `null` - is handled via a call to `jQuery.find.attr( elem, name )` which is `Sizzle.attr( elem, name )`, that defers to `Sizzle.selectors.attrHandle`, exposed to jQuery via `jQuery.expr.attrHandle` and that, in turn, is set for boolean attrs by jQuery.

timmywil (approved):

mgol:
@timmywil I introduced two changes here in a fixup commit:
1. The `extraBoolAttrValues` check now lowercases the tested value.
2. Tests for `extraBoolAttrValues` are added. They are only supported when running with jQuery 4.x, but I skipped the warning when running with 3.x (you can see it in added tests) for two reasons:
    1. I didn't want to warn against setting a valid value for which jQuery 4.0.0 adds support.
    2. If someone is setting the `hidden` value to `"until-found"` and that gets normalized to `"hidden"`, that's a bug that an upgrade to jQuery 4.x will likely fix.

Let me know what you think.

timmywil (approved):

mgol:
Landed on both `main` & `3.x-stable` in 54711094c87f50f7dde83d7d6add012ac10cc9fa; the `main` patch is tweaked in #555 and the `3.x-stable` one in #554.

PR #551, opened 2024-10-29
Build: Test more browsers and jquery versions in pull requests

mgol:
Improve the actions setup for PRs so that more issues are detected pre-merge.

Changes:
1. Test on Edge in IE mode
2. Test on more jQuery versions: `git`, `3.x-git`, `3.7.1` & their slim versions as well as full versions of each `3.x` jQuery minor.

This is still fewer tests than post-merge: we don't test all slim versions and Migrate dev, but this should cover the majority of cases.

This was triggered by me breaking `main` despite all my PRs being green when merging. Migrate often treats older Core versions differently so testing on at least all full minors is pretty important...


mgol:
Of course, this will fail now since `main` is broken. That's a good thing. This PR can be reviewed now, but should only be merged once `main` is fixed and this PR rebased & green.

package.json
```diff
@@ -26,6 +26,7 @@
 		"pretest": "npm run npmcopy && npm run build && npm run lint",
 		"start": "npm run npmcopy && node build/tasks/build-watch.js",
 		"test:browser": "npm run pretest && npm run test:unit -- -b chrome -b firefox -h",
+		"test:ie": "npm run pretest && npm run test:unit -- -v -b ie",
```
timmywil, line 29:
```suggestion
		"test:ie": "npm run pretest && npm run test:unit -- -b ie -h",
```

Because this is Edge, I think this can be headless.

.github/workflows/browsers-local.yml
```diff
@@ -0,0 +1,125 @@
+name: Browsers - local
```
timmywil, line 1:
```suggestion
name: Browser Tests
```

It may not be Browserstack, but I wouldn't call it local either.

.github/workflows/browsers-local.yml
```diff
@@ -0,0 +1,125 @@
...
+  NODE_VERSION: 20.x
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: ${{ matrix.MIGRATE_VERSION }} (Chrome/Firefox)
+    strategy:
+      fail-fast: false
+      matrix:
+        MIGRATE_VERSION: ["min"]
```
timmywil, line 16:
```suggestion
        MIGRATE_VERSION: ["min", "esmodules"]
```

.github/workflows/browsers-local.yml
```diff
@@ -0,0 +1,125 @@
...
+          key: ${{ runner.os }}-node-${{ env.NODE_VERSION }}-npm-lock-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-${{ env.NODE_VERSION }}-npm-lock-
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run tests
+        run: |
+          npm run test:browser -- \
```
timmywil, line 41: We can make this easier to read with separate jobs for Chrome and Firefox. Set the browser using matrix values and use `test:unit` instead.

.github/workflows/browsers-local.yml
```diff
@@ -0,0 +1,125 @@
...
+            --jquery 3.3.1 --jquery 3.2.1 --jquery 3.1.1 --jquery 3.0.0 \
+            --retries 3 --hard-retries 1
+
+  ie:
+    runs-on: windows-latest
+    name: ${{ matrix.MIGRATE_VERSION }} (Edge in IE mode)
+    strategy:
+      fail-fast: false
+      matrix:
+        MIGRATE_VERSION: [ "min" ]
```
timmywil, line 56: No need for matrix. If you want to set it in one place, use env instead.

timmywil (changes requested):
Technically, we could run most tests on Safari or Windows instead of Ubuntu and shorten the config by 1/3.

Also, please make your commit message more specific. For instance,

> Build: Test more browsers and jquery versions in pull requests

.github/workflows/browsers-local.yml
mgol: Hmm, but we do also test `esmodules`? Doesn't that conflict with each other?

mgol:
> Technically, we could run most tests on Safari or Windows instead of Ubuntu and shorten the config by 1/3.

I was following what we have set up in Core.

I think Ubuntu actions run faster in CI than Windows - see this UI one for an example - so it's best to use them when Windows or macOS is not needed.

timmywil:
Indeed, I edited my comment. Ubuntu is worth keeping.

.github/workflows/browsers-local.yml
timmywil: We only test `esmodules` in the ubuntu job.

timmywil (approved):

.github/workflows/browsers-local.yml
mgol: Changed

.github/workflows/browsers-local.yml
mgol: Changed

.github/workflows/browsers-local.yml
```diff
@@ -0,0 +1,125 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: ${{ matrix.MIGRATE_VERSION }} (Chrome/Firefox)
+    strategy:
+      fail-fast: false
+      matrix:
+        MIGRATE_VERSION: ["min"]
+        include:
+          - MIGRATE_VERSION: "esmodules"
```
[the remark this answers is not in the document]
mgol: Changed

.github/workflows/browsers-local.yml
mgol: Changed

.github/workflows/browsers-local.yml
mgol: Changed

mgol:
@timmywil this is now blocked on landing the fixes, i.e. PR #552, please review that one.

PR #552, opened 2024-10-29
Core: Fix the selector module, make CI green

mgol:
~~**Note:** Please don't review the first commit, that's #551. I based it like that so that I can see the tests pass; it should be "unbased" before merging. For this reason alone I'm marking this as a draft.~~

Changes:
1. Support legacy pseudos only in jQuery 3.7.0+.
2. Downgrade sinon to a version working with IE 9+.
3. Skip cross-domain ajax tests in IE 9.
4. Skip tests requiring `jQuery.Deferred.getErrorHook` in <3.7.
5. Don't require warnings for backwards-compatible pseudos if Proxy unsupported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
 		"qunit": "2.21.0",
 		"rollup": "4.22.4",
 		"selenium-webdriver": "4.21.0",
-		"sinon": "9.2.4",
+		"sinon": "7.5.0",
 		"uglify-js": "3.9.4",
 		"yargs": "17.7.2"
 	},
--- a/src/jquery/selector.js
+++ b/src/jquery/selector.js
@@ -1,3 +1,4 @@
+import { jQueryVersionSince } from "../compareVersions.js";
 import { migratePatchFunc, migrateWarnProp, migrateWarn } from "../main.js";
 
 // Now jQuery.expr.pseudos is the standard incantation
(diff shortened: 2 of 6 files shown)


mgol:
Since Edge in IE mode only covers IE 11, I also run the following locally on this branch and it passed:
```
npm run test:unit -- \
  -v --browserstack "IE_10" --browserstack "IE_9" \
  --jquery-migrate dev \
  --jquery 3.x-git --jquery 3.x-git.slim \
  --jquery 3.7.1 --jquery 3.7.1.slim --jquery 3.6.4 \
  --jquery 3.5.1 --jquery 3.4.1 \
  --jquery 3.3.1 --jquery 3.2.1 \
  --jquery 3.1.1 --jquery 3.0.0
```

mgol:
Reviewing with whitespace changes disabled is much easier, BTW (too much indenting).

test/unit/jquery/selector.js
```diff
@@ -189,11 +189,20 @@ QUnit.test( "custom pseudos", function( assert ) {
...
+QUnit[
+	jQueryVersionSince( "3.7.0" ) ? "test" : "skip"
+]( "backwards-compatible custom pseudos", function( assert ) {
 	assert.expect( 7 );
 
+	var expectWarningWithProxy = typeof Proxy !== "undefined" ?
+		expectWarning :
+		function( _assert, _title, fn ) {
+			fn();
+			assert.ok( true, "No Proxy => warnings not expected" );
```
mgol, line 201: This is added so that the number of assertions is identical in both versions.

package.json
```diff
@@ -58,7 +58,7 @@
 		"qunit": "2.21.0",
 		"rollup": "4.22.4",
 		"selenium-webdriver": "4.21.0",
-		"sinon": "9.2.4",
```
timmywil, line 60: I almost said something in the previous PR, but forgot the main branch still supports IE9.

src/jquery/selector.js
```diff
@@ -1,3 +1,4 @@
+import { jQueryVersionSince } from "../compareVersions.js";
```
timmywil, line 1: Does this import mean we don't need it as a global in the eslint config?

timmywil (approved):

src/jquery/selector.js
mgol: Imports are in source, globals are in tests; the ESLint global is for tests. Tests are loaded as regular scripts.

mgol: ...which annoys me a lot, as WebStorm keeps inserting the imports the first time I add usage in a test file and I have to remove it. 😅 When we no longer support IE, we can consider loading test files as modules and remove all the globals (plus, this would allow us to only keep one version of these utils; now they're duplicated).

PR #554, opened 2024-11-04
Core: Remove support for jQuery 4.x

mgol:

PR #559, opened 2025-01-10
Data: Patch camelCase behavior of $.fn.data, warn about Object.prototype

mgol:
~~NOTE: Right now, this PR includes PR #554; only review the commit titled "Data: Patch camelCase behavior of $.fn.data, warn about Object.prototype". This PR should not be merged before #554; that's the only reason it's in a draft state.~~

Changes:
1. Patch not only `jQuery.data()`, but also `jQuery.fn.data()`.
2. Patch `jQuery.removeData()` & `jQuery.fn.removeData()` to work in most cases
   when different keys with the same camelCase representation were passed to
   the data setter and later to `removeData`.
3. Warn about using properties inherited from `Object.prototype` on data
   objects.

### DIFF
--- a/src/jquery/data.js
+++ b/src/jquery/data.js
@@ -1,44 +1,344 @@
 import { migratePatchFunc, migrateWarn } from "../main.js";
 import { camelCase } from "../utils.js";
 
-var origData = jQuery.data;
+var rmultiDash = /[A-Z]/g,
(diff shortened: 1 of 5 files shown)


mgol:
Tests are passing on all tested browsers, BTW: https://github.com/mgol/jquery-migrate/actions/runs/12713854358/job/35442704508

timmywil (approved):

PR #563, opened 2025-02-18
Build: upgrade dependencies, including jtr@0.2.5

timmywil:
- does not update sinon, uglify-js, or eslint


mgol (approved):
LGTM, but updating UglifyJS - and Simon, though not to the latest version - should be fine on `main`, shouldn't it?

timmywil:
Good point. Updated as much as I could.

mgol (approved):

Issue #524, opened 2024-07-15
Provide exports setup similar to jQuery 4.0.0


mgol:
See https://github.com/jquery/jquery/wiki/jQuery-4-exports-explainer

The Migrate setup would likely be simpler, maybe we can avoid the factory mode, there's no slim mode, etc., but some of that would still translate back here.

mgol:
PR: #566

PR #566, opened 2025-03-31
Core: Provide a package.json exports setup inspired by jQuery Core

mgol:
The Migrate setup is simpler than Core as it doesn't have the slim or factory versions, but the core ideas are similar.

Fixes gh-524
Ref jquery/jquery#5255
Ref jquery/jquery#5429


timmywil (approved):
LGTM

PR #571, opened 2025-04-08
Core: Make migrateWarnProp not return the prop for disabled patches

mgol:
Also:
1. Don't warn just on access of `jQuery.fn.push`, etc., but when it's called.
2. Rename some internal methods.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -70,38 +70,71 @@ export function migrateInfo( code, msg ) {
 	migrateMessageInternal( code, msg, "info" );
 }
 
-function migrateMessagePropInternal(
+function migratePatchPropInternal(
 	obj, prop, value, code, msg, migrateMessageFn
 ) {
+	var orig = obj[ prop ];
 	Object.defineProperty( obj, prop, {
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -111,34 +144,28 @@ function migrateMessageFuncInternal(
 			migrateMessageFn( code, msg );
 		}
 
-		// Since patches can be disabled & enabled dynamically, we
-		// need to decide which implementation to run on each invocation.
-		finalFunc = jQuery.migrateIsPatchEnabled( code ) ?
-			newFunc :
-
-			// The function may not have existed originally so we need a fallback.
-			( origFunc || jQuery.noop );
```
mgol, line 120: I'm not sure why I added this `noop` fallback originally, but if the patch is disabled, the function should not be available if it wasn't there originally. I added tests for that (the `push` & `isArray` ones).

mgol:
Adding "Discuss in Meeting" as this blocks another patch for 4.0.0.

timmywil (approved):

PR #577, opened 2025-04-28
Deferred: Don't warn on setting getStackHook to the getErrorHook value

mgol:
Don't warn if `jQuery.Deferred.getStackHook` is set to the value of `jQuery.Deferred.getErrorHook`. This is to facilitate plugins supporting both jQuery <3.7 and older without triggering Migrate warnings and without requiring complex logic parsing `jQuery.fn.jquery`.


mgol:
For reference, this is what you currently have to do to avoid Migrate messages if you want to define hook in all versions where it's supported and you support both jQuery <3.7 and older:
```js
function getErrorHook() { /* ... */ }

var version = jQuery.fn.jquery
	.split( "." )
	.map( function( v ) {
		return Number( v );
	} );

if ( version[ 0 ] >= 4 || ( version[ 0 ] === 3 && version[ 1 ] >= 7 ) ) {
	jQuery.Deferred.getErrorHook = getErrorHook;
} else {
	jQuery.Deferred.getStackHook = getErrorHook;
}
```

timmywil (approved):

PR #588, opened 2025-08-11
Docs: Improve the solution for `boolean-attributes`

mgol:
Because jQuery 4.0.0 will not transform attribute values in the getter, the following code changes between jQuery 3.x & 4.x:

```js
$( elem ).prop( "disabled", true );
$( elem ).attr( "disabled" );
```

In jQuery 3.x, the reported attribute value is `"disabled"`, in jQuery 4.x it will be an empty string. This may create issues when migrating so Migrate patches and warns against it.

Add a recommendation to avoid using the `.attr()` getter on boolean attrs.


timmywil (approved):

mgol:
Landed on `main` in e61af3f8e108f3816e439693c12c23caeaed12c5 and on `3.x-stable` in 1436296b73056bb5d37f2b0ee0275bc490f2fd52.

